PR #1, opened 2022-12-22
Update the basic Python scaffold

BrunoGrandePhD:
The goal of this PR is to adapt the package skeleton provided by PyScaffold for `dcqc`. Some of these changes were copied from `sagetasks`, which was also packaged using PyScaffold. Here are the main changes:

- Use CodeCov instead of Coveralls for tracking code coverage
  - I was having issues when I tried using Coveralls with `sagetasks`
- Add `mypy` (static type checker) as a pre-commit hook
- Create a `Pipfile` that installs the package and its dependencies
- Update the README and CONTRIBUTING docs

PR #3, opened 2023-01-07
Implement `SynapseFS`, a file system-like abstraction for file operations

BrunoGrandePhD:
The goal of this PR is to provide an implementation of the PyFileSystem `FS` abstract class for Synapse called `SynapseFS`. The `FS` class is reminiscent of the `FileSystem` interface in Java, which we were using for the `nf-synapse` Nextflow plugin. A neat thing about PyFileSystem is that it comes with an extensive suite of unit tests. 

There are many projects at Sage that could potentially benefit from this implementation (_e.g._ schematic, nf-synstage, SynapseFormation). One of the amazing benefits of using `SynapseFS` is that you can swap it with a fast in-memory file system during testing. As long as you use `fs.open_fs()`, you can swap a `syn://` URL for a `mem://` URL. 

In the long term, I think we can consider implementing other file systems such as SevenBridges and Terra. This way, our tooling can use a common interface for all of these storage backends, making us incredibly flexible as a team. 

Right now, the SynapseFS tests take a while since they're all integration tests that actually make calls to Synapse. Ideally, if we think this approach is viable, we would move the SynapseFS code to a separate repository so that (1) multiple projects can pull it in as a dependency and (2) the long integration tests would be done separately. 

Depends on #1


src/dcqc/filesystems/synapsefs.py
```diff
@@ -0,0 +1,640 @@
...
+        return parent_id
+
+    def _get_parent_entity(self, entity_id: str) -> str:
+        entity = self.synapse.get(entity_id, downloadFile=False)
+        if isinstance(entity, Project):
+            message = f"Project ({entity.id}) has no parent."
+            raise ValueError(message)
+        parent_id = entity.parentId
+        parent = self.synapse.get(parent_id, downloadFile=False)
+        return parent.id
```
thomasyu888, line 275: You are not returning the parent entity here, but only it's id.  Should this just return `parent_id` or should the function be renamed?

tests/test_filesystems.py
```diff
@@ -0,0 +1,344 @@
...
+    def make_fs(self):
+        folder_name = self.id()
+        self.folder = Folder(folder_name, parent=self.test_root)
+        self.folder = self.synapse.store(self.folder)
+        synapse_fs = SynapseFS(self.folder.id, self.auth_token)
+        return synapse_fs
+
+    def destroy_fs(self, fs):
+        fs.close()
+        self.synapse.delete(self.folder)
```
thomasyu888, line 174: I'm assuming that `make_fs` is what assigns the `self.fs`?  Or does this come from the extension of the `FSTestCases`.

If I'm understanding correctly, you are instantiating a FS that has the root folder `self.folder.id` - Is that correct?

src/dcqc/filesystems/synapsefs.py
```diff
@@ -0,0 +1,640 @@
...
+            children_list = self._get_children(current_entity)
+            if self.is_synapse_id(part):
+                children = {entity["id"]: entity for entity in children_list}
+            else:
+                children = {entity["name"]: entity for entity in children_list}
+
+            if part in children:
+                current_entity = children[part]["id"]
+            else:
+                raise ResourceNotFound(path)
```
thomasyu888, line 212: Nit: These lines of code took me a couple read throughs to fully understand.  It makes sense why you would want to traverse through each part to get the path to get the synapse id.

src/dcqc/filesystems/synapsefs.py
```diff
@@ -0,0 +1,640 @@
...
+        current_entity = starting_entity
+        for part in path.strip(self.DELIMITER).split(self.DELIMITER):
+            if part in {"", self.DELIMITER, self.CWD_SYMBOL}:
+                continue
+
+            if part == self.PARENT_DIR:
+                current_entity = self._get_parent_entity(current_entity)
+                continue
+
+            children_list = self._get_children(current_entity)
```
thomasyu888, line 203: Just to confirm, when `part` is at its final element, current entity should be the synapse id of the container right before it.  For instance:

```
my/example/here
```

When `part` is `here`, `current_entity` should be the Synapse id of `example` right?

src/dcqc/filesystems/synapsefs.py
```diff
@@ -0,0 +1,640 @@
...
+            Entity: The associated Synapse entity
+
+        Raises:
+            fs.errors.ResourceNotFound: If ``synapse_id`` does not exist.
+            fs.errors.ResourceInvalid: If ``synapse_id`` does not correspond
+                 to a supported entity type.
+        """
+        entity = self.synapse.get(synapse_id, downloadFile=download_file)
+        valid_types = list(self.SUPPORTED_TYPES.values())
+        if not any(isinstance(entity, type_) for type_ in valid_types):
```
thomasyu888, line 233: I think you can do:
```suggestion
        valid_types = tuple(self.SUPPORTED_TYPES.values())
        if not isinstance(entity, valid_types):
```

src/dcqc/filesystems/synapsefs.py
```diff
@@ -0,0 +1,640 @@
...
+                    "content_md5": entity._file_handle.contentMd5,
+                    "version_label": entity.versionLabel,
+                    "version_number": entity.versionNumber,
+                }
+                raw_info["synapse"].update(file_info)
+
+        if "annotations" in namespaces:
+            raw_info["annotations"] = entity.annotations
+
+        return Info(raw_info)
```
thomasyu888, line 375: what does `Info` here do?  Does it just take kwargs?

thomasyu888 (changes requested):
This is a great PR - some comments/changes.  I really like the fact that we can have a common interface to interact with different file systems.

Definitely one of the more complex PR's I have reviewed - awesome!

tests/test_filesystems.py
```diff
@@ -0,0 +1,344 @@
...
+
+
+@pytest.mark.integration
+def test_that_synapsefs_can_be_initialized_with_different_roots():
+    # Rootless
+    SynapseFS()
+    SynapseFS("")
+
+    # Synapse ID for a project or folder
+    SynapseFS("syn50545516")
```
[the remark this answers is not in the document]
BrunoGrandePhD: Done

tests/test_filesystems.py
```diff
@@ -0,0 +1,344 @@
...
+        remote_file = RemoteFile(temp_file, "foo", "w", on_close=None)
+        remote_file.close()
+
+
+def test_that_staging_a_local_file_creates_a_copy(get_data):
+    path = get_data("test.txt")
+    local_fs = open_fs(f"osfs://{path.parent}")
+    with TemporaryDirectory() as tmp_dir_name:
+        tmp_dir_path = Path(tmp_dir_name)
+        target_path = tmp_dir_path / "test.txt"
```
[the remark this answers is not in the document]
BrunoGrandePhD: Indeed, the `Path` class should only be used for local paths. Note that the class will dynamically use `WindowsPath` or `PosixPath` depending on the OS. In your example, I'm guessing that it's interpreting the `s3` as a Windows drive, even if you're not on Windows. 

When using PyFileSystem URIs, I've been using `PurePosixPath`. I can do this because PyFileSystem expects POSIX-style paths even on Windows ([docs](https://docs.pyfilesystem.org/en/latest/concepts.html#paths)). 

The [pathlib docs](https://docs.python.org/3.11/library/pathlib.html#module-pathlib) do a good job at describing that pure paths never access the file system whereas concrete paths (like `Path`) do. In the case of URIs, pure paths make more sense since I typically just want to perform string operations on paths (like joining parts or retrieving the parent).

tests/test_filesystems.py
```diff
@@ -0,0 +1,344 @@
...
+        assert not target_path.exists()
+        target_file = target_path.open("wb")
+        local_fs.download(path.name, target_file)
+        target_file.close()
+        assert target_path.exists()
+
+
+@pytest.mark.integration
+def test_that_staging_a_synapse_file_creates_a_copy(mocker):
+    mocked_synapse = mocker.patch("synapseclient.Synapse")
```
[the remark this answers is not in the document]
BrunoGrandePhD: Not by definition. I mostly added the tag because the test is slow even after mocking Synapse. Rather than adding a `slow` mark for a single test, I just added a comment to clarify my intent here.

src/dcqc/filesystems/synapsefs.py
```diff
@@ -0,0 +1,640 @@
...
+from synapseclient.entity import Entity, File, Folder, Project, is_container
+
+from dcqc.filesystems.remote_file import RemoteFile
+
+if TYPE_CHECKING:
+    from fs.info import RawInfo
+
+
+@contextmanager
+def synapse_errors(path):
```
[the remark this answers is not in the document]
BrunoGrandePhD: The `FSTestCases` suite checks if certain `fs` errors are raised in appropriate circumstances. So, `synapse_errors()` is a context manager for mapping `synapseclient` errors to equivalent `fs` errors. This works towards a unified interface so clients can intercept `fs` errors and handle them however they want without knowing anything about the Synapse Python client.

src/dcqc/filesystems/synapsefs.py
```diff
@@ -0,0 +1,640 @@
...
+
+        Returns:
+            Synapse: Authenticated Synapse client
+        """
+        if not hasattr(self._local, "synapse"):
+            synapse = Synapse(**self.synapse_args)
+            synapse.login(authToken=self.auth_token)
+            self._local.synapse = synapse
+        # Clear the Synapse cache to ensure up-to-date
+        self._local.synapse.cache.purge(after_date=0)
```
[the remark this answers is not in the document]
BrunoGrandePhD: That's intentional. I was running into some strange behaviors when the caching mechanism was enabled. I don't think it plays well with SynapseFS, which tries to manage the lifecycle of files with the `RemoteFile` class. 

That said, this would clear the cache on any computer from regular `synapseclient` usage, which is arguably an unwanted side effect. I'll try setting the cache to a new (and thus empty) temporary directory. It won't be used because I typically specify the download location of files, but at least the Synapse client won't use information from previous `synapseclient` work.

src/dcqc/filesystems/synapsefs.py
BrunoGrandePhD: Good catch! I changed the return value from the entity to the ID at the last minute, but I forgot to update the method name.

tests/test_filesystems.py
BrunoGrandePhD: `make_fs()` is called by `setUp()` in the base test class [here](https://github.com/PyFilesystem/pyfilesystem2/blob/8ed9dc495d8ba2f83fbb2a1145d34d92e13644be/fs/test.py#L268-L269), which in turn is called automatically by `unittest` for each test method. 

And you're correct on the second point. I'm creating a new Synapse folder to act as the root for each individual test. It's part of why this test suite takes so long, but it ensures that the tests don't run into concurrency problems.

src/dcqc/filesystems/synapsefs.py
BrunoGrandePhD: Good catch!

src/dcqc/filesystems/synapsefs.py
BrunoGrandePhD: You can explore the class [here](https://github.com/PyFilesystem/pyfilesystem2/blob/8ed9dc495d8ba2f83fbb2a1145d34d92e13644be/fs/info.py#L30), but it's essentially a wrapper around the `raw_info` dictionary with several convenience functions. You can review [these docs](https://docs.pyfilesystem.org/en/latest/info.html) for more details on to use the `Info` class.

src/dcqc/filesystems/synapsefs.py
BrunoGrandePhD: On the `self._get_children(current_entity)` line, that's correct. The `current_entity` will get updated with the Synapse ID corresponding to `part` a little further below on the `children[next_part]["id"]` line. 

That said, I agree that the fact that `current_entity` and `part` don't track together is confusing. I've renamed `part` to `next_part` so they don't seem like they should be in sync.

src/dcqc/filesystems/synapsefs.py
BrunoGrandePhD: This function is critical to SynapseFS, so I've added comments to explain what it does.

src/dcqc/filesystems/synapsefs.py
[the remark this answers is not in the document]
BrunoGrandePhD: I confirmed that caching is indeed the root cause. When caching is enabled, you can see that version 2 below is identical (3 bytes, _i.e._ `bar`) despite the local file containing `barbaz`. After deleting the `.cacheMap` file, I re-run the `synapse.store()` and you can see that version 3 now has 6 bytes as expected. 

As a compromise, I'm going to create temporary SynapseFS-specific cache directories and also purge them to avoid caching issues since I don't see another way of clearing the cache. This way, the user's usual cache (`~/.synapseCache`) is unaffected by SynapseFS. Best of both worlds. 

image

src/dcqc/filesystems/synapsefs.py
```diff
@@ -0,0 +1,666 @@
...
+            root (Optional[str]): _description_
+
+        Raises:
+            CreateFailed: If the root is not or does not start with a Synapse ID.
+            CreateFailed: If the root does not resolve to a project or folder.
+
+        Returns:
+            Optional[str]: A single Synapse ID to act as the root.
+        """
+        if root is None or root == "":
```
BWMac, line 144: What is the importance of allowing the file system to be rootless or not? Why might someone choose to use this one way vs. the other? Just trying to wrap my head around this part

src/dcqc/filesystems/synapsefs.py
```diff
@@ -0,0 +1,666 @@
...
+        if isinstance(entity, Project):
+            message = f"Project ({entity.id}) has no parent."
+            raise ValueError(message)
+        parent_id = entity.parentId
+        with synapse_errors(parent_id):
+            parent = self.synapse.get(parent_id, downloadFile=False)
+        return parent.id
+
+    def _get_children(self, entity_id: str) -> list[dict]:
+        """_summary_
```
BWMac, line 303: NIT: missing summary

tests/conftest.py
```diff
@@ -7,4 +7,30 @@
...
+
+CNFPATH = Path(__file__).resolve()
+TESTDIR = CNFPATH.parent
+DATADIR = TESTDIR / "data"
+
+UUID = str(uuid4())
+
+
+def pytest_configure():
+    pytest.UUID = UUID  # type: ignore
```
BWMac, line 23: This UUID becomes the name of the S3 bucket directory containing info about the test run, right? Is there a reason to use something like a UUID instead of a descriptive string with info about the test run (including a datetime of the run or other info about the testing)?

BWMac (commented):
This was really interesting to review - a lot of stuff I haven't seen before that I still need to spend time learning about. Just a few comments and questions.

src/dcqc/filesystems/synapsefs.py
BrunoGrandePhD: Great question! 

The main use case for rootless SynapseFS is the ability to open files and folders wherever they are (as long as your Synapse account has access to them). If you provide a root (_e.g._ a Synapse project), you can't open anything outside of that root. In the case of DCQC, the Synapse URIs might point anywhere, so it's preferable to not specify a root. 

Another reason is the user experience (UX) for opening files. In DCQC, all I know about a file is its URI, such as `syn://syn50555279`. If I didn't have rootless mode (which was the case initially), how would I initialize `SynapseFS` with? I eventually added support for opening a root like `syn://syn50555279/..` (_i.e._ open the given file's parent directory). However, this is clunky and unintuitive. 

Overall, I concluded that adding a rootless mode resulted in a more elegant UX: 

```python
# Without rootless mode
fs = SynapseFS("syn://syn50555279/..")
with fs.open("syn50555279") as infile:
    contents = infile.read()

# With rootless mode
fs = SynapseFS()
with fs.open("syn50555279") as infile:
    contents = infile.read()
```

tests/conftest.py
BrunoGrandePhD: Another good question! 

The UUIDs are used to create a single folder on Synapse, which will contain one folder for each test. Usually, the tests are run sequentially, and the `setUp()` and `tearDown()` methods will result in the folder being created and deleted before the next test is run. 

Originally, I was using `time.time_ns()` (current time since epoch in nanoseconds) to uniquely identify folders that I was creating. I occasionally ran into some weird issues that I presumed were related to concurrency, so I switched to a statistically unique test run ID using `uuid4()`. I haven't had the weird issues since. 

That said, I could include additional information about the test run in the name alongside the UUID to get the best of both worlds.

src/dcqc/filesystems/synapsefs.py
thomasyu888: Thanks for linking the docs.  Makes sense - looks like there are some nice class methods you are able to use in `Info`

tests/test_filesystems.py
```diff
@@ -0,0 +1,341 @@
...
+    with pytest.raises(CreateFailed):
+        SynapseFS("DCQC Test Project")
+
+    # Path that doesn't start with a Synapse ID
+    with pytest.raises(CreateFailed):
+        SynapseFS("DCQC Test Project/syn50557597")
+
+
+@pytest.mark.integration
+def test_that_a_rootless_synapsefs_can_open_a_random_file_by_id(synapse_fs):
```
thomasyu888, line 50: Ah, I missed this test - I made a comment in another PR asking about this.  You can ignore that comment since rootless synapse fs can open a random file by id.

Just out of curiosity, what happens when you pass in a Folder or Project?

BrunoGrandePhD: As long as you're using an operation that works with files and folders (like `listdir()`), it works. But if you pass a folder to `open()`, you'll get an error indicating that a file was expected.

tests/conftest.py
thomasyu888: +1

src/dcqc/filesystems/synapsefs.py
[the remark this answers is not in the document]
thomasyu888: Awesome - we may want to document this particular design decision somewhere.

I read through both `IOBase` [docs](https://docs.python.org/3/library/io.html#io.IOBase) and the `RemoteFile` class, can you talk about how they are managing the lifecycle of files?

Now I'm curious on figuring out what causes this behavior in the python client (_in the future_), nothing in the `on_close` function in this PR makes it seem like the store function won't upload a new version.

thomasyu888 (approved):
@BrunoGrandePhD I'm going to approve, but did you want to create a PR that merges "PyScaffold-Tweaks" into main first?

This way you won't have a huge PR after everything goes into "PyScaffold-Tweaks"

src/dcqc/filesystems/synapsefs.py
BWMac: This makes a lot of sense, thanks for the response!

src/dcqc/filesystems/synapsefs.py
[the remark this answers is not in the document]
BrunoGrandePhD: I added a link to this discussion where I purge the cache for context. 

There's something in the Synapse Python client that's causing it to upload the same unmodified file as the second version instead of picking up on the modified file and uploading that instead. 

When I say that RemoteFile is managing the lifecycle of the file, I'm referring to the fact that once the file operation is done (_e.g._ `open()` with a context manager), the file is uploaded and then deleted. This might break assumptions being made by the caching mechanism in the Synapse Python client.

BrunoGrandePhD:
The `PyScaffold-Tweaks` branch is for posterity. I already merged #1 into `main`. I just updated the base branch for this PR to `main`. 

The issue was caused by having `:` in the date, which was causing issues since Synapse doesn't support colons in folder names.

PR #4, opened 2023-01-07
Implement `FileType`, `File`, and `Target` to representing the targets of quality control

BrunoGrandePhD:
The goal of this PR is to implement the low-level layers when it comes to representing the subject of quality control, namely files. The `Target` class has changed a lot since the proof of concept and is now very simple. In fact, as it stands, it can be renamed to `MultiFile` or just `FileSet` since it's just a wrapper around a list of files. Before deleting this class, I want to implement a multi-file QC test (_e.g._ for paired FASTQ files) to see if the `Target` class remains useful. 

The lower-level `File` class is composed of a URL (which can be local paths) and file metadata (which includes the file type). The `file_type` key in the metadata is used to pull up an instance of the `FileType` data class, which acts as a container of file type–related information, like file extensions. In the proof of concept, this class also took care of download the files, but this is now delegated to PyFileSystem using `open_fs()` on the file URLs. 

This PR also introduces the `SerializableMixin` class. This is the first time that I use a mixin like this. It acts like an interface, and it helps with type checking. As I demonstrated in the proof of concept, many of these objects need to be serializable either to be passed to Nextflow or for generating JSON reports. 

Depends on #3


src/dcqc/file.py
```diff
@@ -0,0 +1,125 @@
+from __future__ import annotations
+
+import os
+import re
+from collections.abc import Collection, Mapping
+from copy import deepcopy
+from dataclasses import asdict, dataclass
```
thomasyu888, line 7: TIL: dataclasses is a standard python library since Python 3.7.

src/dcqc/utils.py
```diff
@@ -0,0 +1,25 @@
...
+    scheme, separator, path = url.rpartition("://")
+    if separator == "":
+        prefix = "osfs://"
+    else:
+        prefix = scheme + separator
+
+    # parent_path can be "" if there is no "/" in the path
+    parent_path, _, base_name = path.rpartition("/")
+    parent_url = prefix + parent_path
+    fs = open_fs(parent_url)
```
thomasyu888, line 24: Just to confirm if you wanted to download a file: `File("syn://syn50555279").stage()`,

1. This creates a rootless `fs`
2.  rootless fs can actually download a file entity without specifying the full path: `syn://syn50545516/test.txt`

Or am I not understanding this?

tests/conftest.py
```diff
@@ -25,12 +30,50 @@ def pytest_configure():
...
         return path
 
     yield _get_data
+
+
+outputs = set()
+
+
+@pytest.fixture
+def get_output():
```
thomasyu888, line 46: How is this function being used?

thomasyu888 (changes requested):
Looks great!

src/dcqc/mixins.py
```diff
@@ -0,0 +1,15 @@
+from abc import ABC, abstractmethod
+from typing import Union
+
+SerializedObject = dict[str, Union[str, int]]
+
+
```
thomasyu888, line 6: I say that this is the appropriate way to provide this feature in a lot of classes.

src/dcqc/file.py
```diff
@@ -0,0 +1,125 @@
...
+    # picked up by @dataclass as an instance attribute
+    _registry = dict()  # type: ignore
+
+    # Instance attributes
+    name: str
+    file_extensions: tuple[str, ...]
+
+    def __init__(self, name: str, file_extensions: Collection[str]):
+        self.name = name
+        self.file_extensions = tuple(file_extensions)
```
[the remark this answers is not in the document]
BrunoGrandePhD: Here, I had to override the default `__init__()` method because I needed to run `register_file_type()` as part of initialization. Also, since the `str.endswith()` method only works with tuples, I wanted to cast any sequence of strings to a tuple. The different in type hints is that internally, the `file_extensions` attribute will be stored as a tuple, but you can provide any sequence of file extensions to the class.

tests/test_file.py
```diff
@@ -0,0 +1,70 @@
...
+def test_that_a_remote_file_is_staged_when_requesting_a_local_path():
+    url = "syn://syn50555279"
+    metadata = {"file_type": "txt"}
+    file = File(url, metadata)
+    assert not file.is_local()
+    file.get_local_path()
+    assert file.is_local()
+
+
+def test_that_a_local_file_is_not_moved_when_requesting_a_local_path(test_files):
```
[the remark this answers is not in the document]
BrunoGrandePhD: Exactly. You can think of this test as a requirement:

> As a DCQC user, I expect that local files are not moved or copied when they are QC'ed. 

If anything changes in `File` or anything that's lower-level like SynapseFS that causes this requirement to no longer being met, we'll know right away with the test suite. Then, we'll at least have the opportunity to decide whether to change the requirement (and communicate the change with our users) or fix the implementation.

src/dcqc/utils.py
thomasyu888: Resolving this, because this was included as a test case.

src/dcqc/file.py
```diff
@@ -0,0 +1,125 @@
...
+        file_type = file_type.lower()
+        if file_type not in cls._registry:
+            types = list(cls._registry)
+            message = f"File type ({file_type}) not among available options ({types})."
+            raise ValueError(message)
+        return cls._registry[file_type]
+
+
+# Instantiated file types are automatically tracked by the FileType class
+FileType("*", ())  # To represent all file types
```
thomasyu888, line 50: I'm ok with this, but what are your thoughts on a configuration file that we can add and remove from to track these FileType classes? 

Perhaps add a TODO if you think thats a good idea?

src/dcqc/file.py
```diff
@@ -0,0 +1,125 @@
...
+        self,
+        url: str,
+        metadata: Mapping[str, Any],
+        relative_to: Optional[Path] = None,
+    ):
+        if relative_to is not None and self.is_local(url):
+            scheme, separator, resource = url.rpartition("://")
+            path = Path(resource)
+            if not path.is_absolute():
+                resource = os.path.relpath(relative_to / resource)
```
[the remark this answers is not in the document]
BrunoGrandePhD: In that case, it's assumed to be relative to the current work directory. I've added the following line to make this explicit:

```python
relative_to = relative_to or Path.cwd()
```

src/dcqc/utils.py
BrunoGrandePhD: 1. It doesn't necessarily create a rootless FS. It tries to open an FS for the parent directory, and if one cannot be inferred from the URL (_e.g._ `syn://syn50555279`), then it opens a rootless FS. 
2. In the case of a rootless SynapseFS (I can't speak to other FS classes yet), you don't need the full path. You can either refer to the file directly with its Synapse ID, or you can refer to it by name by starting a path with a Synapse ID of one of its parent folders/project and navigating to the file.

tests/conftest.py
BrunoGrandePhD: I created it for one of the tests, but I ended up switching to another approach. I've deleted this function for now.

src/dcqc/file.py
BrunoGrandePhD: The idea of a "database" of file types (_e.g._ with a CSV file) also crossed my mind. I just didn't see any benefits for the time being. Can you think of any? 

While it would allow users to add their own file types, they wouldn't be able to take advantage of them unless they also created a suite. Also, they can already add new file types by instantiating the `FileType` class; they're automatically tracked wherever they are created. 

I did add a TODO to track the idea in case something changes and makes the change worthwhile.

thomasyu888: Ah, thats what I was missing - users need to create a suite to take advantage of their own filetypes.  Not sure how feasible this is, but one use case I can think of for a "database" of filetypes is if we can encapsulate the information required for both `FileType` and `Suite`.

thomasyu888 (approved):
LGTM!

src/dcqc/file.py
BrunoGrandePhD: Yup, I have [another TODO](https://github.com/Sage-Bionetworks-Workflows/py-dcqc/blob/0ddb5837795b7ece8afd90246bb4ec4034e1a0b0/src/dcqc/suites/suites.py#L6-L7) that would have the file types also include the associated with the suites. We would have to figure out a way to ensure that a dependency isn't introduced from file types to suites, which would break the layered architecture.

PR #5, opened 2023-01-07
Implement `TestABC` and `SuiteABC` as well as subclasses for the testing logic

BrunoGrandePhD:
The goal of this PR is to implement the QC testing logic in the form of tests and suites, which are sets of tests for a given target. For now, you can consider targets to be the same as files since we're only handling single-file targets. 

In both cases, the base classes for tests and suites are abstract classes from the `abc` module. This allows you to mark methods as abstract methods that need to be implemented by all subclasses. For internal tests, this is the `compute_status()` method, and for external tests, it's the `generate_process()` method. Currently, the `SuiteABC` class doesn't have an abstract method, so I might remove `ABC` as its base class. 

I'm using another mixin class for external tests, `ExternalTestMixin`. This test overrides the `compute_status()` method and delegates that logic to the `_interpret_process_outputs()` method, which defaults to checking the process exit code. As you can see, this results in minimal logic being necessary when implementing both internal and external tests. 

I'm also introducing a small `Process` data class, which will contain the container and command information. I've also added memory and CPU attributes. You'll note that the command is now being stored as a list of strings similar to how `subprocess` handles commands. I'm using the `shlex` module to generate the full command string. This minimizes any risk of running into issues with shell quoting. 

Depends on #4


src/dcqc/suites/suite_abc.py
```diff
@@ -0,0 +1,157 @@
...
+        superclasses = cls.__mro__
+        superclasses = reversed(superclasses)
+        all_tests = []
+        for cls in superclasses:
+            if hasattr(cls, "add_tests"):
+                add_tests = set(cls.add_tests)  # type: ignore
+                all_tests.extend(add_tests)
+            if hasattr(cls, "del_tests"):
+                del_tests = set(cls.del_tests)  # type: ignore
+                all_tests = [test for test in all_tests if test not in del_tests]
```
thomasyu888, line 61: Nit: If all tests are a set as well you can do

```
all_tests.difference(del_tests)
```

src/dcqc/suites/suite_abc.py
```diff
@@ -0,0 +1,157 @@
...
+        self.required_tests = set(required_tests).intersection(test_names)
+
+        skipped_tests = skipped_tests or list()
+        self.skipped_tests = set(skipped_tests).intersection(test_names)
+
+        self.tests = self.init_test_classes()
+        self._status = TestStatus.NONE
+
+    @classmethod
+    def list_test_classes(cls):
```
thomasyu888, line 51: Nit: Missing doc strings & type hinting

Lets think about which functions should return a list, which should a set.

src/dcqc/suites/suite_abc.py
```diff
@@ -0,0 +1,157 @@
...
+                all_tests.extend(add_tests)
+            if hasattr(cls, "del_tests"):
+                del_tests = set(cls.del_tests)  # type: ignore
+                all_tests = [test for test in all_tests if test not in del_tests]
+        return all_tests
+
+    @classmethod
+    def _default_required_tests(cls) -> list[str]:
+        test_classes = cls.list_test_classes()
+        required_tests = filter(lambda test: test.tier <= 2, test_classes)
```
thomasyu888, line 67: Should probably document that by default, tier 1 and 2 tests are required

src/dcqc/suites/suite_abc.py
```diff
@@ -0,0 +1,157 @@
...
+        return suite_status
+
+    def to_dict(self):
+        suite_status = self.compute_status()
+        test_dicts = []
+        for test in self.tests:
+            test_dict = test.to_dict()
+            test_dict.pop("target", None)  # Remove redundant `target` info
+            test_dicts.append(test_dict)
+        suite_dict = {
```
thomasyu888, line 148: Should this structure be captured in a python class in the future?

src/dcqc/tests/tests.py
```diff
@@ -0,0 +1,99 @@
+import hashlib
+
+from dcqc.enums import TestStatus
+from dcqc.file import File
+from dcqc.tests.test_abc import ExternalTestMixin, Process, TestABC
+
+
+class FileExtensionTest(TestABC):
```
thomasyu888, line 8: This is great for now, but I can imagine this module exploding into thousands of lines.  Any ideas for scaling?

tests/test_suites.py
```diff
@@ -0,0 +1,77 @@
...
+
+def test_that_the_ome_tiff_suite_has_a_superset_of_the_tiff_suite_tests():
+    tiff_tests = TiffSuite.list_test_classes()
+    ome_tiff_tests = OmeTiffSuite.list_test_classes()
+    assert set(ome_tiff_tests) > set(tiff_tests)
+
+
+def test_that_a_test_suite_can_be_retrieved_by_name():
+    actual = SuiteABC.get_suite_by_name("OmeTiffSuite")
+    assert actual is OmeTiffSuite
```
thomasyu888, line 44: We discussed this on the call, but the reason `is` is used because these are not instantiated yet so they are not `isinstance`.

thomasyu888 (changes requested):
🚀   Just some comments/questions.  Excited to see all the tests - I will agree that it's slightly confusing to have the `tests` directory for developer tests and then  the `tests` directory for all the tests, but this structure makes sense.

src/dcqc/suites/suite_abc.py
BrunoGrandePhD: Good idea!

src/dcqc/tests/tests.py
BrunoGrandePhD: I'm not sure yet. I almost went with one submodule per tier, but that didn't feel right since it's arbitrary. One submodule file per test seems excessive. I figured that I would decide once we have more tests in front of us, which should hopefully help with deciding how to organize them.

src/dcqc/suites/suite_abc.py
BrunoGrandePhD: We could follow the [Memento design pattern](https://refactoring.guru/design-patterns/memento), which uses a class to capture the instance state. In this case, I opted against it (for now) because these dictionaries are only mainly used for serialization. 

If they prove to be burdensome to maintain, we could look into creating simple `dataclasses` for passing this information along.

BrunoGrandePhD:
I enabled the following options for `mypy`, which will flag any untyped definitions in `dcqc`. Just as a heads-up, my latest commit does have a lot of these type hints filled in. 

```
disallow_untyped_calls = true
disallow_untyped_defs = true
disallow_incomplete_defs = true
check_untyped_defs = true
```

src/dcqc/suites/suite_abc.py
```diff
@@ -70,10 +70,10 @@ def list_test_classes(cls) -> tuple[Type[TestABC], ...]:
         for cls in reversed(superclasses):  # Start from the base class
             if hasattr(cls, "add_tests"):
                 add_tests = set(cls.add_tests)  # type: ignore
-                all_tests.union(add_tests)
+                all_tests |= add_tests
```
thomasyu888, line 73: TIL: these short cuts

src/dcqc/suites/suite_abc.py
thomasyu888: Thats an interesting design pattern you bring up - I think this is great for now.

Were you thinking that you would use the memento design pattern to track the different validation statuses of a target?

thomasyu888 (approved):
I'm going to approve this , just one small question which doesn't block the merging of this PR.

src/dcqc/suites/suite_abc.py
BrunoGrandePhD: I didn't like having to reassign the output of `union()` to the variable. 

```python
all_tests = all_tests.union(add_tests)
```

That said, I found the `*update()` set functions, so I'm using that instead. 

```python
if hasattr(cls, "add_tests"):
    add_tests = set(cls.add_tests)  # type: ignore
    all_tests.update(add_tests)
if hasattr(cls, "del_tests"):
    del_tests = set(cls.del_tests)  # type: ignore
    all_tests.difference_update(del_tests)
```

BrunoGrandePhD: I'll make this change in the following PR.

thomasyu888: TIL: `difference_update`

PR #6, opened 2023-01-07
[ORCA-90] Implement missing functionality for a user acceptance test

BrunoGrandePhD:
The goal of this PR is to implement any missing functionality to achieve a reasonable user acceptance test. The full test is described in `tests/test_acceptance.py`, but I've included a summary below. You can download the report [here](https://www.synapse.org/#!Synapse:syn50696454). 

```python
parser = CsvParser("files.csv")
suites = parser.create_suites(required_tests, skipped_tests)
report = JsonReport("syn://syn50696438/report.json")
report.save(suites)
```

As you can see in the above snippet, I had to implement the following pieces of code: 

- A CSV parser that can generate lists of files, targets, and suites from a CSV file
- A JSON report generator that can create reports and save them anywhere using PyFileSystem

Depends on #5

### DIFF
--- a/src/dcqc/parsers.py
+++ b/src/dcqc/parsers.py
@@ -1,7 +1,7 @@
 import csv
-import os
-from collections.abc import Iterator
+from collections.abc import Collection, Iterator
 from pathlib import Path
+from typing import Optional
 
 from dcqc.file import File
 from dcqc.suites.suite_abc import SuiteABC
@@ -21,33 +21,26 @@ class CsvParser:
                 yield index, row
 
     def _row_to_file(self, row: dict[str, str]) -> File:
-        csv_path = self.path
+        csv_directory = self.path.parent
         url = row.pop("url")
-        if File.LOCAL_REGEX.match(url):
-            scheme, separator, resource = url.rpartition("://")
-            path = Path(resource)
-            if not path.is_absolute():
-                resource = os.path.relpath(csv_path.parent / resource)
-            url = "".join([scheme, separator, resource])
-        file = File(url, row)
+        file = File(url, row, relative_to=csv_directory)
         return file
 
-    def list_targets(self) -> Iterator[Target]:
+    def create_targets(self) -> Iterator[Target]:
         for _, row in self.list_rows():
             file = self._row_to_file(row)
             yield Target(file)
 
-    def list_suites(self) -> Iterator[SuiteABC]:
-        # TODO: Generalize this function to support multi-file targets
-        for target in self.list_targets():
-            if len(target.files) > 1:
-                message = (
-                    f"Target ({target}) is composed of more than one file. "
-                    "`iter_suites()` currently only supports single-file targets."
-                )
-                raise ValueError(message)
+    def create_suites(
+        self,
+        required_tests: Optional[Collection[str]] = None,
+        skipped_tests: Optional[Collection[str]] = None,
+    ) -> Iterator[SuiteABC]:
+        for target in self.create_targets():
+            # This function assumes that there is one file per target,
+            # which is always the case when parsing CSV files (for now)
             first_file = target.files[0]
             file_type = first_file.get_file_type()
             suite_cls = SuiteABC.get_suite_by_file_type(file_type)
-            suite = suite_cls(target)
+            suite = suite_cls(target, required_tests, skipped_tests)
             yield suite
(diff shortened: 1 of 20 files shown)


tests/data/files.csv
```diff
@@ -1,3 +1,6 @@
 url,file_type,md5_checksum
 test.txt,TXT,14758f1afd44c09b7992073ccf00b43d
-test.txt,TXT,definitelynottherightmd5checksum
+test.txt,TIFF,14758f1afd44c09b7992073ccf00b43d
```
thomasyu888, line 3: Nit: I see that file_type is all capitalized, should this be case agnostic when we are actually using it to determine the tests to run per file?

src/dcqc/parsers.py
```diff
@@ -0,0 +1,50 @@
...
+from collections.abc import Collection, Iterator
+from pathlib import Path
+from typing import Optional
+
+from dcqc.file import File
+from dcqc.suites.suite_abc import SuiteABC
+from dcqc.target import Target
+
+
+class CsvParser:
```
thomasyu888, line 11: I like this a lot.  In general, I think that most of the modules you have contributed in this suite of PR's could have docstrings so that people can go to the module and have an understanding on a high level of the classes and functions the module contains.

That said, you have done a wonderful job of naming each of your modules.

thomasyu888 (approved):
Awesome!  This PR is pretty self explanatory, so I'm approving!

src/dcqc/parsers.py
```diff
@@ -0,0 +1,50 @@
+import csv
```
thomasyu888, line 1: In general, I think that most of the modules you have contributed in this suite of PR's could have docstrings so that people can go to the module and have an understanding on a high level of the classes and functions the module contains.

That said, you have done a wonderful job of naming each of your modules.

tests/test_acceptance.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+@pytest.mark.integration
+def test_json_report_generation(get_data):
+    # GIVEN a list of external tests to skip (to remain self-contained)
+    all_tests = TestABC.list_tests()
+    skipped_tests = [test.__name__ for test in all_tests if test.is_external_test]
+
+    # AND a subset of internal tests to be required (to verify suite status behavior)
+    required_tests = ["Md5ChecksumTest"]
```
thomasyu888, line 18: Would we want the specification of required tests to be part of `files.csv`?

src/dcqc/reports.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __init__(self, url: str, overwrite: bool = False) -> None:
+        self.url = url
+        self._fs, self._path = open_parent_fs(url)
+        if self._fs.exists(self._path) and not overwrite:
+            message = f"URL ({url}) already exists. Set `overwrite=True` to ignore."
+            raise FileExistsError(message)
+
+    def to_file(self, obj: Any):
+        with self._fs.open(self._path, "w") as outfile:
```
thomasyu888, line 20: I really like that upon closing done by the context manager, it uploads the file to Synapse if you specify a path in Synapse.  Incredible.

tests/data/files.csv
BrunoGrandePhD: Comparisons are made after making everything lowercase ([source](https://github.com/Sage-Bionetworks-Workflows/py-dcqc/blob/b21007d1fc57515abb1e0ef7845e66c23a86c425/src/dcqc/file.py#L43)).

src/dcqc/parsers.py
BrunoGrandePhD: It depends on need and alternative implementations. For example, instead of implementing a `FileViewParser`, we would add support for file views to SynapseFS, which can return them as CSV files. In this case, we could just re-use CsvParser.

src/dcqc/parsers.py
BrunoGrandePhD: I added some docstrings to classes in #5. Now that the package structure is settled, I feel better about adding docstrings without having to throw them away.

tests/test_acceptance.py
BrunoGrandePhD: When I add the CLI, I will have to think of how I want to expose these settings. One drawback of including required or skipped tests in the CSV file is that if these are file-type specific, there will be a lot of redundant information. 

Another option is a YAML configuration file, which would allow for non-redundant filetype/test-specific configuration, but I need to think about it some more.

thomasyu888: You bring up good points.  I'm sure we will run into this in the near future, so we can discuss this then.  No need to decide right now.

PR #7, opened 2023-01-24
[IBCDPE-457] Improve and document `file` submodule

BrunoGrandePhD:
This PR was something I started working on as a side-project, but I ended up building on top of these improvements for IBCDPE-448, so I'm including it here as a PR related to a subtask. Note that I originally intended on filling in docstrings for all submodules, but just doing it for the `file` submodule took a lot of time because I also made improvements to its API. My goal is to progressively fill in docstrings as I work on py-dcqc. 

Here's a summary of the changes introduced here:

- I documented the `file` submodule. 
  - Most of the changes in `file.py` are documentation-related. 
  - I noticed a few low-hanging fruits for improvements.
- New `interrogate` pre-commit hook.
  - This will help maintain "docstring coverage" once we get that percentage up.
  - Note that the minimum allowed is intentionally low to allow the pre-commit hooks to pass. 
  - We will steadily increase this number as we document the package. 
- I made some improvements to how the reference docs are generated
  - I'm not done polishing this process. 
  - The `sphinx_autodoc_typehints` package uses the type hints in the auto-generated docs. 
  - This ensures that the type info isn't duplicated between type hints and the docstrings. 
  - In other words, you no longer have to document the types under `Args:` or `Returns:`.
- I made several tweaks to type hints in general. 
  - These can probably be ignored since they are inconsequential. 
  - Note that the `mypy` pre-commit hook ensures that everything is working as far as typing goes.
- While documenting the `File` class, I noticed that I could improve the way that it used the `SerializableMixin`.
  - Since I like using `Path` objects from `pathlib`, I've added a `dict_factory` function that converts these to strings during the export to dictionaries. 
  - This ensures that the dictionary is indeed JSON-serializable. 
  - This way, I can still use the `dataclass.asdict()` function.
- I discovered `typing.ClassVar` for differentiating class attributes from instance attributes. 
  - This is really handy for dataclasses because type-hinted attributes are automatically used for constructing default magic methods. 
  - By annotating an attribute with `ClassVar`, the `@dataclass` decorator ignores those attributes. 
- I moved the `TestStatus` enum into the `test_abc` submodule. 
  - It looked lonely in the `enum` submodule on its own. 
  - I might change this back if I start accumulating more enums.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,4 +53,10 @@ repos:
 - repo: https://github.com/kynan/nbstripout
   rev: '0.6.1'
   hooks:
-    - id: nbstripout
+  - id: nbstripout
+
+- repo: https://github.com/econchick/interrogate
+  rev: 1.5.0
+  hooks:
+  - id: interrogate
+    args: [--verbose, --fail-under=25]
--- a/src/dcqc/file.py
+++ b/src/dcqc/file.py
@@ -1,13 +1,25 @@
+"""Represent local and remote files and their metadata.
+
+See module dcqc.target for the multi-file target class.
+
+Classes:
+
+    FileType: For collecting file type-specific information.
+    File: For bundling file location and metadata as well as
+    operations for retrieving file contents.
+"""
+
 from __future__ import annotations
 
 import os
 import re
 from collections.abc import Collection, Mapping
 from copy import deepcopy
-from dataclasses import asdict, dataclass
+from dataclasses import dataclass
 from pathlib import Path
 from tempfile import mkdtemp
-from typing import Any, Optional
+from typing import Any, ClassVar, Optional
+from warnings import warn
 
 from fs.base import FS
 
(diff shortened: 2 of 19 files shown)


.pre-commit-config.yaml
```diff
@@ -53,4 +53,10 @@ repos:
 - repo: https://github.com/kynan/nbstripout
   rev: '0.6.1'
   hooks:
-    - id: nbstripout
+  - id: nbstripout
+
+- repo: https://github.com/econchick/interrogate
```
thomasyu888, line 58: TIL: This

src/dcqc/file.py
```diff
@@ -96,93 +131,187 @@ def __init__(
 
         self.local_path = local_path or self._init_local_path()
 
+    def __hash__(self):
+        return hash((self.url, self.type, tuple(self.metadata.items())))
+
+    def __eq__(self, other):
+        return hash(self) == hash(other)
```
thomasyu888, line 138: Great additions

src/dcqc/file.py
```diff
@@ -96,93 +131,187 @@ def __init__(
...
+        A destination is not required for remote files; it
+        defaults to a temporary directory.
         Local files aren't moved if a destination is omitted.
 
         Args:
-            destination (Optional[str]): File or folder path
-                where to store the file. Defaults to None.
-            overwrite (bool): Whether to ignore existing file
-                at the target destination. Defaults to False.
+            destination: File or folder where to store the file.
```
thomasyu888, line 311: Did you create a custom template for the autoDocGenerator via vscode? https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring

thomasyu888 (approved):
Great improvements and documentation.  I liked many of the code simplification and additions. 🛸

thomasyu888:
A side note - I think this is a wonderful example of how developer tests enabled you to improve the codebase without having to update the tests but make sure the code still meets all the requirements.  Fantastic.

src/dcqc/file.py
BrunoGrandePhD: I was already using this extension, but I didn't know that it supported custom templates. Sweet! 

image

BrunoGrandePhD:
> a wonderful example of how developer tests enabled you to improve the codebase without having to update the tests but make sure the code still meets all the requirements.

@thomasyu888: I'm glad that you noticed that!

PR #8, opened 2023-01-24
[IBCDPE-458] Improve APIs in general for CLI

BrunoGrandePhD:
In practice, I made these changes as I was implementing the CLI, but I want to dedicate a PR to everything excluding to CLI to make it easier for review. 

Here's a summary of what changed here:

- Implement the `JsonParser` class
  - This class is the magic that can construct (_i.e._ deserialize) an object that's been serialized to JSON
- Improve `JsonReport` class
  - I noticed that this class could be improved to be a more general "JSON exporter"
  - Notably, I added the `save_many()` method to export a series of objects to separate JSON files
  - I might update the name later
  - I didn't want to update it now because it would make it hard to split up the PRs
- I have added some methods for reconstructing objects
  - Most of these are called `from_dict()`, but the `TestABC` class has some additional constructors as class methods
- I have added some methods to `Target` that currently operate on its files
- I have updated the `open_parent_fs()` utility function to be more flexible
  - Before, it would open the immediate parent folder of a given file as an FS
  - Now, it opens to "highest available" folder in the path as an FS
  - The main difference here is that the immediate parent folder doesn't need to exist anymore
  - The calling function can run `fs.makedirs()` to fill in the nested directories

Depends on #7


src/dcqc/suites/suite_abc.py
```diff
@@ -171,3 +232,37 @@ def to_dict(self) -> SerializedObject:
             "tests": test_dicts,
         }
         return suite_dict
+
+    @classmethod
+    def from_dict(cls, dictionary: SerializedObject) -> SuiteABC:
```
thomasyu888, line 237: Nit: These `from_dict` functions would fail with a `KeyError` error if a dictionary with missing keys was passed in.  Not that this is a huge deal, but just some food for thought.  Should we be concerned about those scenarios?

src/dcqc/utils.py
```diff
@@ -11,7 +11,12 @@ def open_parent_fs(url: str) -> tuple[FS, str]:
         prefix = scheme + separator
 
     # parent_path can be "" if there is no "/" in the path
-    parent_path, _, base_name = path.rpartition("/")
+    parent_path, _, base_name = path.partition("/")
```
thomasyu888, line 14:
```
path = "my/path/to/parent/basename"
path.partition("/")
# ('my', '/', 'path/to/parent/basename')
```

This is being pedantic, but my brain is so used to `base_name` being `basename`.  Can you speak to the update here? (No need to update the variable names - I'm just curious)

demos/DCQC.ipynb
```diff
@@ -133,8 +133,8 @@
    "metadata": {},
    "outputs": [],
```
thomasyu888, line 134: I'm running through the demo, and I'm getting:

```
AttributeError: type object 'TestABC' has no attribute 'list_tests'
```

demos/DCQC.ipynb
```diff
@@ -133,8 +133,8 @@
...
    "source": [
-    "report = JsonReport(report_url, overwrite=True)\n",
-    "saved = report.save(suites)"
+    "report = JsonReport()\n",
```
thomasyu888, line 136: This causes an error:

```
ValueError: '/Users/tyu/sage/py-dcqc/tests/data/test.txt' is not in the subpath of '/Users/tyu/sage/py-dcqc/demos' OR one path is relative and the other is absolute.
```

thomasyu888 (changes requested):
Nice changes - No major comments, with the exception that the demo notebook doesn't run all the way through for me.

demos/DCQC.ipynb
BWMac: I'm getting the same error, is it because the line: `all_tests = TestABC.list_tests()` should be `all_tests = TestABC.list_subclasses()`?

BWMac (commented):
The demo doesn't run all the way for me either, even with swapping `list_tests()` for `list_subclasses()`, due to the URL error that Tom commented on. I really like the idea of including these demos though. Makes the functionality easily accessible.

src/dcqc/parsers.py
```diff
@@ -45,10 +52,83 @@ def create_suites(
...
+            raise ValueError(message)
+
+        return object_
+
+    def load_json(self) -> Any:
+        with self.path.open("r") as infile:
+            contents = json.load(infile)
+        return contents
+
+    def get_class(self, cls_name: str) -> Type[SerializableMixin]:
```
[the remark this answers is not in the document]
BrunoGrandePhD: There is a difference. 

First, `Type` and `TypeVar` (used to create `T`) mean different things. You can check out the `typing` docs for more information, but the examples below will demonstrate their usage. 

Second, the `Type[SomeClass]` annotation informs the static type checker to expect `SomeClass` (sub)class objects instead of `SomeClass` (sub)class instances. 

Third, when you generate a type variable using `TypeVar` (_e.g._ `T`, which is a conventional shorthand), you can use this value to inform the type checker that its value is consistent throughout the function signature. I just added a function [here](https://github.com/Sage-Bionetworks-Workflows/py-dcqc/blob/df99b50df1b2620aebbd496d7f67ed67cf8bc56a/src/dcqc/mixins.py#L68-L79) that uses a type variable (relevant snippet copied below). In this example, the function signature is interpreted by the type checker as follows: the return type will be the same as the type of `self`, and this type must be a instance of `SerializableMixin`. 

```
T = TypeVar("T", bound="SerializableMixin")

def copy(self: T) -> T:
    ...
```

The reason that I can't write `def copy(self: SerializableMixin) -> SerializableMixin:` is that this is interpreted as follows: the return type will be an instance of `SerializableMixin`, and the type of `self` will be an instance of `SerializableMixin`, but there is no relationship between the two types. In other words, this doesn't convey to the type checker that the return type will match the input type. 

Finally, in the case of the `JsonParser`, I have the following method signature. Given that `JsonParser` can produce objects from any `SerializableMixin` subclass, the type checker didn't like the having the return type being `SerializableMixin` because that isn't specific enough. For example, if I use a `TestABC`-specific method (_e.g._ `get_status()`) after deserializing a test using `JsonParser`, the type checker thinks that's an error because the return value is `SerializableMixin`, which doesn't have a `get_status()` method. I wrote the `parse_expected()` method to overcome this, which also enables with runtime type checking. The type checker interprets the following method signature as follows: the return type will be an instance of the class defined by `expected_cls`. 

```python
def parse_expected(cls, path: Path, expected_cls: Type[T]) -> T:
    ...
```

src/dcqc/parsers.py
```diff
@@ -45,10 +52,83 @@ def create_suites(
...
+
+        Raises:
+            ValueError: If the JSON file does not
+                describe a expected object.
+
+        Returns:
+            The reconstructed object.
+        """
+        parser = JsonParser(path)
+        object_ = parser.parse_object()
```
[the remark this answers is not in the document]
BrunoGrandePhD: Good point. `parse_expected()` came afterwards. I refactored the class so that `parse_object()` and `parse_objects()` both have an `expected_cls` parameter and removed the `parse_expected()` method.

src/dcqc/suites/suite_abc.py
BrunoGrandePhD: A `KeyError` would be appropriate in that situation. In theory, these dictionaries should only be generated by dcqc's `to_dict()` methods. We should also have tests that ensure that we can perform a serialize/deserialize round-trip without any issues. For now, I don't think we need to support deserialization of arbitrary JSON files.

src/dcqc/utils.py
BrunoGrandePhD: I added a comment and updated the variable names to clarify this change. For example, `basename` is better called `path` to use the FS terminology. 

```
    # Retrieve the "top-most" parent folder for the FS root
    # The remainder of the string can be used as the FS path
    fs_root, _, path = resource.partition("/")
```

demos/DCQC.ipynb
BrunoGrandePhD: I've decided to delete this file because it duplicates the `test_json_report_generation` test, which passes on my end. I might add it back once I have a more tutorial-like structure that I would like to follow. 

```
tests/test_acceptance.py::test_json_report_generation PASSED
```

demos/DCQC.ipynb
BrunoGrandePhD: See above comment.

BrunoGrandePhD:
As I mentioned in the comments, I opted to delete the `DCQC.ipynb` file because it duplicated the `test_json_report_generation` test. Once I have a better "user flow" to demonstrate, I'll add a Jupyter notebook.

src/dcqc/parsers.py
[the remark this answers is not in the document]
thomasyu888: Thanks for explanation!

demos/DCQC.ipynb
thomasyu888: Makes sense for this ticket. We will need to have a ticket to create examples to enable ease of use for our stakeholders.   Are you thinking of including examples in the sphinx autodocumentation?

thomasyu888 (approved):
I'm going to approve this - instead of a juypter notebook, we can also potentially include the examples via auto documentation with sphinx.  👍🏼 

It's important that we are able to provide examples to our stakeholders in the coming month so I can also ship this around to get feedback and advertise this (when ready)

BrunoGrandePhD:
According to [this page](https://docs.codecov.com/docs/github-checks), you can hide the CodeCov annotations using the `a` keyboard shortcut (or `i` to hide review comments).

BrunoGrandePhD:
I'm going to ignore the failing coverage checks because some of those gaps in coverage will be filled by the tests added in #9.

PR #9, opened 2023-01-24
[IBCDPE-459] Implement command-line interface (CLI) for dcqc

BrunoGrandePhD:
Alright, this is the "goal PR" for IBCDPE-448, although I've tagged the PR using the relevant subtask. 

In this PR, I'm using `typer` to build a CLI that is roughly equivalent to the one I created for the [proof of concept script](https://github.com/Sage-Bionetworks-Workflows/nf-dcqc-poc/blob/5a849c93ec17341749fbdb0fd7309ba2ca7de630/bin/dcqc.py#L17-L84). I think it went really well! It's also easy to test `typer` commands. 

Each command is document, so it should be self-explanatory. I also included some sample input/output files in the `tests/data/` directory to (1) demonstrate what they look like and (2) for testing purposes.

Depends on #8

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,4 +10,4 @@ dcqc = {editable = true, path = "."}
 dcqc = {editable = true, path = ".", extras = ["testing", "dev"]}
 
 [requires]
-python_version = "3.10"
+python_version = "3.11"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -125,6 +125,14 @@
             "markers": "python_full_version >= '3.6.0'",
             "version": "==3.0.1"
         },
+        "click": {
+            "hashes": [
+                "sha256:7682dc8afb30297001674575ea00d1814d808d6a36af415a82bd481d37ba7b8e",
+                "sha256:bb4d8133cb15a609f44e8213d9b391b0809795062913b383c62be0ee95b1db48"
+            ],
+            "markers": "python_version >= '3.7'",
+            "version": "==8.1.3"
+        },
         "dcqc": {
             "editable": true,
             "path": "."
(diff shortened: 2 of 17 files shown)


thomasyu888 (changes requested):
🔥 This all looks great - I took a look at your tests and install locally and ran everything.

```
dcqc --help
dcqc create-targets tests/data/small.csv create_target
dcqc stage-target tests/data/target.json stage_target

dcqc create-tests -r Md5ChecksumTest tests/data/target.json create_tests
dcqc create-process tests/data/test.external.json create_process/process.json

dcqc compute-test tests/data/test.internal.json compute_test/test.json


dcqc create-suite create_suite/suite.json tests/data/test.computed.json tests/data/test.computed.json tests/data/test.computed.json
# create-suite fails if the create_suite dir doesn't exist, but the rest doesn't fail

dcqc combine-suites combine/suite.json tests/data/suite.json  tests/data/suite.json  tests/data/suite.json
# same here combine doesn't exist
```
Couple of things:

1. I think the missing thing here is the steps written out.  For example, if someone were to develop or use this package locally, I imagine they would want to know what the steps were to perform tests on a particular file.  Can that be added in the readme?
2. `create-suite` and `combine-suites` break the behavior that you've set for the other commands.  The other commands create the folders for you while those commands don't.  We may want to standardize.  What was your rationale in the other commands creating folders?

BrunoGrandePhD:
1. The primary use case for these CLI commands is the Nextflow workflow, which requires smaller steps to enable parallelism. Once we have a use case for using dcqc directly (_i.e._ outside of Nextflow)—which I expect to surface eventually—I can implement a high-level CLI that's user-friendly and document it in the README. I captured this plan in the README. 

2. The `create-suite` and `combine-suites` also create parent folders thanks to the `_create_parent_directories()` method in the `JsonReport` class, which uses URLs instead of Paths. I'm unhappy with the inconsistency between paths and URLs. I plan on resolving that discrepancy in a future iteration with either a mixin or an abstraction (_e.g._ a `URL` class).

thomasyu888 (approved):
Thanks! Going to approve, but the README update hasn't been pushed.

BrunoGrandePhD:
Oups, I forgot to commit the change. Done! Merging now.

PR #10, opened 2023-01-26
[ORCA-119] Fix any issues that come up while testing nf-dcqc on Tower

BrunoGrandePhD:
I'm currently testing nf-dcqc on Tower, and I've come across some minor issues with py-dcqc. I'm opening a PR to accumulate any fixes. I'll be using the manual trigger on the `docker.yml` CI workflow to publish commit-specific containers that Tower can use for testing (instead of minting a release for each iteration).

### DIFF
--- a/src/dcqc/file.py
+++ b/src/dcqc/file.py
@@ -365,6 +365,8 @@ class File(SerializableMixin):
     @wraps(SerializableMixin.to_dict)
     def to_dict(self) -> SerializedObject:
         dictionary = super(File, self).to_dict()
+        # Including the file name in the output for better readability for
+        # unstaged remote files with cryptic URLs (e.g., syn://syn98765432)
         dictionary["name"] = self.name
         return dictionary
 
--- a/src/dcqc/file.py
+++ b/src/dcqc/file.py
@@ -15,6 +15,7 @@ import os
 from collections.abc import Collection, Mapping
 from copy import deepcopy
 from dataclasses import dataclass
+from functools import wraps
 from pathlib import Path
 from tempfile import mkdtemp
 from typing import Any, ClassVar, Optional
@@ -360,6 +361,12 @@ class File(SerializableMixin):
         self.local_path = destination
         return destination
 
+    @wraps(SerializableMixin.to_dict)
+    def to_dict(self) -> SerializedObject:
+        dictionary = super(File, self).to_dict()
+        dictionary["name"] = self.name
+        return dictionary
+
     @classmethod
     def from_dict(cls, dictionary: SerializedObject) -> File:
         """Deserialize a dictionary into a file.
@@ -377,5 +384,8 @@ class File(SerializableMixin):
 
         if dictionary["local_path"] is not None:
             dictionary["local_path"] = Path(dictionary["local_path"])
+
+        # Ignore serialized name since it's a dynamically-computed property
+        dictionary.pop("name", None)
 
         return cls(**dictionary)
--- a/src/dcqc/file.py
+++ b/src/dcqc/file.py
@@ -119,11 +119,12 @@ class File(SerializableMixin):
     def __init__(
         self,
         url: str,
-        metadata: Mapping[str, Any],
+        metadata: Optional[Mapping[str, Any]] = None,
         relative_to: Optional[Path] = None,
         local_path: Optional[Path] = None,
     ):
         self.url = self._relativize_url(url, relative_to)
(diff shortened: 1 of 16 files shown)


src/dcqc/parsers.py
```diff
@@ -45,19 +45,16 @@ def create_files(self) -> Iterator[tuple[int, File]]:
                 file.stage(destination, overwrite=True)
             yield index, file
 
-    def create_targets(self, stage_files: bool = True) -> Iterator[Target]:
+    def create_targets(self) -> Iterator[Target]:
         for index, file in self.create_files():
-            if stage_files:
-                file.stage()
             yield Target(file, id=f"{index:04}")
```
thomasyu888, line 50: I didn't notice this last time, but why 04 and not 05 or 03?

src/dcqc/reports.py
```diff
@@ -46,6 +50,12 @@ def to_file(self, obj: Any, url: str, overwrite: bool):
         #       (e.g., relativize them based on output JSON path)
         with fs.open(fs_path, "w") as outfile:
             json.dump(obj, outfile, indent=2)
+            outfile.write("\n")
+
+    def _generate_single(self, item: SerializableMixin) -> SerializedObject:
+        item.serialize_paths_relative_to(self.paths_relative_to)
+        report = item.to_dict()
```
thomasyu888, line 57: Would it make sense to include the `serialize_paths_relative_to` in the `to_dict` function?

src/dcqc/target.py
```diff
@@ -67,12 +67,13 @@ def get_file_type(self) -> FileType:
...
     def stage(
-        self,
-        destination: Optional[Path] = None,
-        overwrite: bool = False,
-    ) -> Iterator[Path]:
+        self, destination: Optional[Path] = None, overwrite: bool = False
+    ) -> list[Path]:
+        paths = list()
         for file in self.files:
-            yield file.stage(destination, overwrite)
```
thomasyu888, line 75: Why the switch from generator?

thomasyu888 (approved):
🔥 Great work pushing this out - just some minor comments / questions so I'm comfortable approving this.

src/dcqc/file.py
```diff
@@ -360,6 +361,12 @@ def stage(
         self.local_path = destination
         return destination
 
+    @wraps(SerializableMixin.to_dict)
+    def to_dict(self) -> SerializedObject:
+        dictionary = super(File, self).to_dict()
+        dictionary["name"] = self.name
```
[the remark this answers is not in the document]
BrunoGrandePhD: I added a comment to clarify:

```
# Including the file name in the output for better readability for
# unstaged remote files with cryptic URLs (e.g., syn://syn98765432)
```

src/dcqc/parsers.py
BrunoGrandePhD: I feel that the chance that there is more than 999 files in a workflow run is non-trivial, but I think it's rare that a workflow run would have more than 9,999 files. And because I'm iterating through the CSV file, I don't know how many rows there are ahead of time.

src/dcqc/reports.py
BrunoGrandePhD: I considered that, but I didn't want to change the method signature in all places where `to_dict()` is defined. As I mentioned in a `TODO`, I think this logic ultimately belongs in the `JsonReport` class (probably going to rename it to `JsonExporter`) as a custom `JSONEncoder` subclass. I think Paths should be converted to strings as late as possible (_i.e._ during JSON serialization and not in `to_dict()`). This was a quick fix, but I'm not super happy with it. I will likely refactor as soon as I touch the `JsonReport` code.

src/dcqc/target.py
BrunoGrandePhD: Because calling `paths = target.stage()` wouldn't stage the files. Instead, it would return an iterator that you would then need to iterate though (_e.g._ using `list(paths)`) to actually stage the files. I don't think this is intuitive. In any case, there is no real benefit to a generator here since it's not as if large amounts of data are being processed in-memory. Also, most targets only have a single file, so it makes even less sense most of the time.

thomasyu888: Makes sense. Thanks!

src/dcqc/reports.py
thomasyu888: 👍🏼

thomasyu888 (approved):

PR #11, opened 2023-01-27
Publish branch-specific containers

BrunoGrandePhD:
When I was working on #10, I had to manually publish a container (`workflow_dispatch`) for the Nextflow workflow to use every time I pushed a commit. This was annoying, so I want to automatically publish containers with every branch. It also resulted in a simpler workflow with fewer `if` conditions.


thomasyu888 (dismissed):
💯   One thing we can try eventually is using the ghcr.  I still don't fully understand if there are any limitations to github applications, but I like ghcr because it is contained within the repository.

BrunoGrandePhD:
I rebased this branch on `main` so I can merge it before #10. I also discovered `docker/metadata-action` [here](https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-docker-hub-and-github-packages), and it's a much more elegant way of generating the list of tags as well as multi-deployment to Docker Hub and GHCR.io.

BrunoGrandePhD:
I just revamped this PR once again. I decided to merge both CI workflows for two reasons. First, I want to be able to use the exact same file that's tested and uploaded to PyPI in the container image. Second, I want to only publish a container image if the tests pass. While there are ways of setting up inter-workflow dependencies (_e.g._ `workflow_run` or `workflow_call`), the first reason pushed me to integrating everything together.

BrunoGrandePhD:
Merging so I can test the new workflow

PR #12, opened 2023-01-27
Demonstrate implementation for new file types and tests

BrunoGrandePhD:
During today's Code Space meeting, I demonstrated how easy it is to implement new tests for new file types. Instead of throwing away that code, I'm opening a PR, which can act as documentation for now if anyone wants to contribute new tests and/or file types. 

Note that I started the `all` "package extra" to capture dependencies needed for running the tests. This way, the basic dependencies are kept minimal. 

```console
pip install dcqc[all]
```

Here's the relevant snippet from the acceptance test report: 

<details>
<summary>JsonLdSuite</summary>
```json
{
  "type": "JsonLdSuite",
  "target": {
    "type": "Target",
    "id": "0006",
    "files": [
      {
        "url": "tests/data/example.jsonld",
        "metadata": {
          "md5_checksum": "56bb5f34da6d6df2ade3ac37e25586b7"
        },
        "type": "JSON-LD",
        "local_path": "tests/data/example.jsonld",
        "name": "example.jsonld"
      }
    ]
  },
  "suite_status": {
    "required_tests": [
      "Md5ChecksumTest"
    ],
    "skipped_tests": [],
    "status": "passed"
  },
  "tests": [
    {
      "type": "JsonLoadTest",
      "tier": 2,
      "is_external_test": false,
      "status": "passed"
    },
    {
      "type": "JsonLdLoadTest",
      "tier": 2,
      "is_external_test": false,
      "status": "passed"
    },
    {
      "type": "Md5ChecksumTest",
      "tier": 1,
      "is_external_test": false,
      "status": "passed"
    },
    {
      "type": "FileExtensionTest",
      "tier": 1,
      "is_external_test": false,
      "status": "passed"
    }
  ]
}
```
</details>

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -64,8 +64,8 @@ exclude =
 # `pip install dcqc[PDF]` like:
 # PDF = ReportLab; RXP
 
-# Dependending for quality control
-qc =
+# Additional dependencies for compute tests
+all =
     rdflib
 
 # Dependencies for testing (used by tox and Pipenv)
--- a/setup.cfg
+++ b/setup.cfg
@@ -66,7 +66,7 @@ exclude =
 
 # Additional dependencies for compute tests
 all =
-    rdflib
+    rdflib~=6.2
 
 # Dependencies for testing (used by tox and Pipenv)
 testing =
--- a/setup.cfg
+++ b/setup.cfg
@@ -64,6 +64,10 @@ exclude =
 # `pip install dcqc[PDF]` like:
 # PDF = ReportLab; RXP
 
+# Dependending for quality control
+qc =
+    rdflib
+
 # Dependencies for testing (used by tox and Pipenv)
 testing =
     setuptools~=65.0
--- a/src/dcqc/tests/tests.py
+++ b/src/dcqc/tests/tests.py
@@ -1,8 +1,6 @@
 import hashlib
 import json
 
-from rdflib import Graph
-
 from dcqc.file import File
 from dcqc.tests.test_abc import ExternalTestMixin, Process, TestABC, TestStatus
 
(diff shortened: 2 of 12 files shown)


thomasyu888:
@BrunoGrandePhD I will review this soon, but this has incredibly high value.  I can imagine this being demo'ed in a sprint review once we have documentation for how to do this.

setup.cfg
```diff
@@ -64,6 +64,10 @@ exclude =
 # `pip install dcqc[PDF]` like:
 # PDF = ReportLab; RXP
 
+# Additional dependencies for compute tests
+all =
```
thomasyu888, line 68: Eventually once this package gets more complex, will we want to split `all` into something like `jsonldtest` ?

src/dcqc/tests/tests.py
```diff
@@ -36,13 +37,62 @@ def compute_status(self) -> TestStatus:
...
 
+class JsonLoadTest(TestABC):
+    tier = 2
+    only_one_file_targets = False
+
+    def compute_status(self) -> TestStatus:
+        status = TestStatus.PASS
+        for file in self.target.files:
+            if not self._can_be_loaded(file):
+                status = TestStatus.FAIL
```
thomasyu888, line 55: Do you think the `_can_be_loaded` test be implemented for all file types?

thomasyu888 (approved):
💣 Just a couple of questions that can probably be made into tickets if we want to make those improvements in the future.

setup.cfg
BrunoGrandePhD: I considered that, but I also don't want one extra per dependency. Once we have the problem of too many dependencies, we can figure out a way of splitting them up. I think this is relatively a non-issue since most people will be using the container image.

src/dcqc/tests/tests.py
BrunoGrandePhD: Once there is a third internal load test that would use the same pattern, I'll add a helper method that can be used by all three use cases. I might use the [template method](https://refactoring.guru/design-patterns/template-method) design pattern.

thomasyu888 (approved):

PR #16, opened 2023-01-30
Speed up py-dcqc test suite

BrunoGrandePhD:
One of the major motivations for #13 is that the test suite would be much faster without the `SynapseFS` tests. This PR takes advantage of the fact that the tests are now relatively fast. For example, there's no need to limit which tests are run on different Python versions and operating systems. I'm now running all tests on all combinations in the `ci.yml` workflow. 

I also took the opportunity to clean up the pytest marks based on Tom's feedback a few PRs back. I now just have a `slow` mark, which is enforced by the `markers` section in `setup.cfg`. Currently, all of my tests take less than 0.03 seconds except for the large acceptance test (which interacts with Synapse) and a CLI test (which spawns two subprocesses). For now, both of these fall under my definition of slow because I like my tests to be almost instant. I also think that the 0.42s CLI test isn't critical enough to include in all local test runs (but it's worth including in the full test suite in `ci.yml`). 

I've set up `pytest` to exclude slow tests by default, but `tox` is configured to override this to include all tests, which is what's used during CI testing. You can also re-include the slow tests by eliminating the mark filter with `pytest -m ""`. 

Aside from that, I added a `remote` entry to the `test_files` fixture. I found a way to create a "remote file" in-memory for much faster tests compared to using Synapse. No mocking was needed! You'll see `test_files["remote"]` a few times, which was me replacing calls to Synapse to in-memory operations. 

I also removed the `TestCLI` class since the only reason it existed was to mark all included tests as `integration`, but I've now removed that mark, so there's no need for a class anymore. Like I mentioned before, I did label one of these tests with `@pytest.mark.slow`. 

```
14.74s call     tests/test_acceptance.py::test_json_report_generation
0.42s call     tests/test_main.py::test_that_the_module_cli_behaves_the_same_as_the_plain_cli
0.02s call     tests/test_main.py::test_create_targets
0.01s call     tests/test_main.py::test_combine_suites
0.01s call     tests/test_main.py::test_create_suite
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,25 +65,18 @@ jobs:
         - macos-latest
         # TODO: Debug the Windows issues
         # - windows-latest
+    env:
+      OS: ${{ matrix.platform }}
+      PYTHON: ${{ matrix.python }}
     runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python }}
-      - name: Retrieve pre-built distribution files
-        uses: actions/download-artifact@v3
+      - uses: actions/download-artifact@v3
         with: {name: python-distribution-files, path: dist/}
-      - name: Run tests (without integration tests)
-        if: matrix.platform != 'ubuntu-latest' || matrix.python != '3.11'
-        env:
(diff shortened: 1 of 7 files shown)


thomasyu888 (approved):
🚀 Thanks for doing this!  Makes sense to me and glad to see that the separation of repos to decrease complexity of this repository and reduce the time it takes to run the tests.

PR #17, opened 2023-01-30
Revamp the way that local paths are handled by the `File` class

BrunoGrandePhD:
This change was prompted by work at closing the code coverage gap for the `file` submodule. I decided to drop the `get_local_path()` method in favor of a Python property like `name`. 

Since I wanted the local path to be included in the serialized `File` JSON output, I wanted to generalized what I had started with the `name` property. That's what motived the `_serialized_properties` class attribute. If you want a property to appear in the output JSON, you should list it in `_serialized_properties` and the `SerializableMixin` will handle the rest. This allowed me to remove the `File`-specific `to_dict()` implementation. 

The changes to the tests are mostly to use this new API (`file.local_path` instead of `file.get_local_path()`) as well as additional tests to close the test coverage gap. The `file` submodule is now 100% covered. 

```
---------- coverage: platform darwin, python 3.11.1-final-0 ----------
Name                           Stmts   Miss Branch BrPart  Cover   Missing
--------------------------------------------------------------------------
src/dcqc/__init__.py               9      0      0      0   100%
src/dcqc/__main__.py               2      0      0      0   100%
src/dcqc/file.py                 160      0     54      0   100%
src/dcqc/main.py                  99      2     34      2    97%   44-45, 56->60
src/dcqc/mixins.py                70      4     36      3    93%   34-35, 41, 63
src/dcqc/parsers.py              100     21     46      6    77%   43-45, 74-76, 89, 97-98, 103-104, 116-117, 125-134
src/dcqc/reports.py               75      9     22      2    89%   24->27, 36-38, 40-41, 64, 68, 82, 88
src/dcqc/suites/__init__.py        0      0      0      0   100%
src/dcqc/suites/suite_abc.py     164      3     67      2    98%   119, 125-126
src/dcqc/suites/suites.py         18      0      0      0   100%
src/dcqc/target.py                43      1     12      1    96%   64
src/dcqc/tests/__init__.py         0      0      0      0   100%
src/dcqc/tests/test_abc.py       127      9     30      0    94%   63, 132-137, 167-171
src/dcqc/tests/tests.py          100      8     25      2    92%   55-56, 65-66, 78-79, 91-92
src/dcqc/utils.py                 20      0      6      0   100%
--------------------------------------------------------------------------
TOTAL                            987     57    332     18    94%
```

### DIFF
--- a/src/dcqc/file.py
+++ b/src/dcqc/file.py
@@ -15,7 +15,6 @@ import os
 from collections.abc import Collection, Mapping
 from copy import deepcopy
 from dataclasses import dataclass
-from functools import wraps
 from pathlib import Path
 from tempfile import mkdtemp
 from typing import Any, ClassVar, Optional
@@ -101,6 +100,7 @@ FileType("TIFF", (".tif", ".tiff"), "format_3591")
 FileType("OME-TIFF", (".ome.tif", ".ome.tiff"), "format_3727")
 
 
+# TODO: Leverage post-init function in dataclasses
 @dataclass
 class File(SerializableMixin):
     """Construct a File object.
@@ -113,10 +113,11 @@ class File(SerializableMixin):
             current work directory (default).
     """
 
+    _serialized_properties = ["name", "local_path"]
+
     url: str
     metadata: dict[str, Any]
     type: str
-    local_path: Optional[Path]
 
     def __init__(
         self,
@@ -136,8 +137,8 @@ class File(SerializableMixin):
         self._fs_path = None
         self._name: Optional[str]
         self._name = None
-
-        self.local_path = local_path or self._init_local_path()
+        self._local_path: Optional[Path]
+        self._local_path = local_path
 
     def __hash__(self):
         return hash((self.url, self.type, tuple(self.metadata.items())))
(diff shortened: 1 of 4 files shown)


src/dcqc/file.py
```diff
@@ -210,6 +196,25 @@ def _init_fs(self) -> tuple[FS, str]:
         self._fs = fs
         return fs, fs_path
 
+    @property
+    def local_path(self) -> Path:
```
thomasyu888, line 200: Nice refactor. Do you think we will need `@property.setters`?

src/dcqc/file.py
```diff
@@ -101,6 +100,7 @@ def get_file_type(cls, file_type: str) -> FileType:
 FileType("OME-TIFF", (".ome.tif", ".ome.tiff"), "format_3727")
 
 
+# TODO: Leverage post-init function in dataclasses
```
thomasyu888, line 103: At some point, all of the todos should be translated to Jira tickets so that the work can be shared across dpe.

thomasyu888 (approved):
🎈 nice! Just some comments, but approving.

src/dcqc/file.py
BrunoGrandePhD: In the case of `local_path` and `name`, I don't want users to be able to change them. In the case of `local_path`, users can change it using `stage()`.

src/dcqc/file.py
BrunoGrandePhD: Most of them don't need to be converted into tickets. Many of these are reminder to refactor bits the next time I touch that code. I would argue that creating tickets will make it less likely for these to get done because everything else will always be higher priority. Instead, I prefer that we be constantly vigilant for improvement opportunities and conduct these improvements as we work on user-driven tickets. You'll come across this point in the Clean Coder book.

PR #21, opened 2023-03-08
Copy image to Quay.io to Docker Hub rate limits

BrunoGrandePhD:
This is a short-term fix to the Docker Hub rate limits that we easily run into due to our current infrastructure.


thomasyu888 (approved):
👍🏼 Thanks for the short term fix!

PR #22, opened 2023-03-08
Add suite and test for rudimentary risk phrase detection (part of ORCA-136)

adamjtaylor:
To help learn how to contribute tests to DCQC, I'm opening this PR to add a test that greps for "date" or "time" in a case-insensitive fashion within files. 

This could be expanded to form a suite (initially called `RiskTermsSuite`) that flags when known strings appear anywhere within files.

This has proved to be a relatively effective and lightweight way to find image acquisition date/timestamps within TIFF/SVS/OME-TIFF and similar files.

This is a WIP and I appreciate advice from @BrunoGrandePhD and others on how to fill out/complete.


adamjtaylor:
I think that I need to add a tests following the pattern of `test_that_the_libtiff_info_test_correctly_interprets_exit_code_0_and_1` and `test_that_the_libtiff_info_test_command_is_produced`

src/dcqc/tests/tests.py
```diff
@@ -142,3 +142,23 @@ def generate_process(self) -> Process:
...
+        command_args = [
+            "grep",
+            "-E",
+            "-w",
+            "-i",
+            "'date|time'",
+            path,
+        ]
+        process = Process(
+            container="quay.io/sagebionetworks/bftools:latest",
```
adamjtaylor, line 161: I guess this can be any generic container?

src/dcqc/suites/suites.py
```diff
@@ -28,3 +28,8 @@ class TiffSuite(FileSuite):
 class OmeTiffSuite(TiffSuite):
     file_type = FileType.get_file_type("OME-TIFF")
     add_tests = (tests.OmeXmlSchemaTest, tests.BioFormatsInfoTest)
+
+class RiskTermsSuite(FileSuite):
```
thomasyu888, line 32: Thanks @adamjtaylor for the contribution!  We haven't officially discussed internally how we would support non-filetype suites.  I do potentially see this to be useful. Just so we can better assist, what was your rationale behind creating this class?

src/dcqc/tests/tests.py
BrunoGrandePhD: Agreed. You can use this one: 

```
quay.io/biocontainers/coreutils:8.30--h14c3975_1000
```

src/dcqc/suites/suites.py
BrunoGrandePhD: This raises an important point. So far, we've focused on file formats like JSON, TIFF, and OME-TIFF. However, there are tests that are less related to a specific format and more related to a "data type". For example, confirming the strandedness of a pair of RNA-seq FASTQ files combines the FASTQ file format with the RNA-seq data type. 

Similarly, this PR introduces (I think) the concept of a plain text file that might contain sensitive information, such as dates or times. Let's imagine a `PhiFree` data type. Thinking out loud here, this kind of check might be adapted based on the file format. For example, in the case of CSV/TSV files, maybe you restrict the test to the headers, whereas you would only check keys in the case of a JSON file. 

All this to say that `dcqc` should be able to handle both concepts (file formats and data types) as well as their interaction (_e.g._ inferring the strandedness for an RNA-seq experiment is different depending on whether you have FASTQ or BAM/CRAM files).

src/dcqc/tests/tests.py
BrunoGrandePhD: Alternatively, we could perform this test in Python (_i.e._ an internal test). For this, we would drop the `ExternalTestMixin` parent class. If this is only intended to be performed on small files, then an internal test is slightly better because you can run it more easily outside of Nextflow.

tests/test_tests.py
```diff
@@ -190,3 +190,32 @@ def test_that_a_process_can_be_serialized_and_deserialized():
     process_dict = process.to_dict()
     process_from_dict = Process.from_dict(process_dict)
     assert process_dict == process_from_dict.to_dict()
+
+def test_that_the_grep_date_test_correctly_interprets_exit_code_0_and_1(
```
BrunoGrandePhD, line 194: I applaud you for writing the test. I hope we can simplify the testing facilities so it's less work to add new tests.

src/dcqc/suites/suites.py
adamjtaylor: Thanks @thomasyu888 and @BrunoGrandePhD!

My rationale for the `RiskTermSuite` class was that it is conceivable that there are other risky strings that we might want to generically detect within any file using grep.

To @BrunoGrandePhD's point I was thinking that tasks within RiskTermSuite might need to be applied to several different file types, and possibly all file types. I concur with the concept of applying  test in different ways depending on the file type.

If easier I would be happy to restrict the suite/test to OME-TIFF or TIFF files. I will also need it to support other imaging formats in the near future including SCN and SVS.

thomasyu888: Thanks @adamjtaylor , this makes sense.   My 2 cents is to add the tests to OME-TIFF and TIFF files for now, which will give us time to think more deeply about how we want to support data type suites.

There is also another use case which is "project specific suites".

adamjtaylor:
Thank you @thomasyu888 for making some minor changes including removing the `RiskTermSuite` and moving the `GrepDateTest` test into the TIFF suite.

We know that there are cases where this test will fail for a false positive (for example in the case of a string like `Date=REDACTED`. I therefore would like to see this test integrated as an `Optional Test` where failing this test does not cause the whole suite to fail. We can link to the relevant tickets on adding this feature

thomasyu888:
Thanks @adamjtaylor for the information.  Pinging @BrunoGrandePhD for visibility.  Due to the current DCQC behavior as outlined in slack (I will be shifting that documentation elsewhere), the way to achieve this would be to do something like this:

```
nextflow run Sage-Bionetworks-Workflows/nf-dcqc \
  -revision dev \
  -profile docker \
  --input testdata/input_txt.csv \
  --required_tests FileExtensionTest,Md5ChecksumTest,LibTiffInfoTest,BioFormatsInfoTest,OmeXmlSchemaTest
```

I do realize this is a bit clunky, but we are working on doing a design of the default behavior which we will circulate soon.

adamjtaylor:
I'm happy to do a feature branch in the sage repo

adamjtaylor:
So if I understand you correctly @thomasyu888 all tests are optional unless included in required tests

thomasyu888:
@adamjtaylor. Thanks for the feedback. 

1. Currently all DCQC tier 1 and tier2 tests are required and ALL tests are executed.
2. Currently, defining "required_tests" means the tests that need to pass for the file to be valid (not whether they are executed). `skipped_tests` means tests to not be executed.
3. For this particular test to be optional (not skipped), the only way to achieve this right now is by specifying all the `--required_tests` as the nextflow parameter, but leaving out the GrepDateTest.  This will still execute the GrepDateTest, BUT it won't impact the overall "validity" of the file.
4. Currently, there isn't a way of specifying the required or skipped tests in the INPUT for DCQC - it would be great to have a discussion about this so we can gather more information.

BrunoGrandePhD:
The current behavior for specifying tests is admittedly confusing. In short, users either globally opt out of a subset of default tests (using `--skipped-tests`) or globally opt into a custom set of tests (using `--required-tests`). Here, "globally" refers to the fact that you cannot require or skip tests per file format. I think this behavior needs some work, and that's the goal of the "DCQC Test Specification" design document. I agree that we need to implement support for "optional" tests that can help surface potential issues without failing QC.

PR #24, opened 2023-03-30
[ORCA-179] Interpret the absence of CLI values for list options as `None` instead of empty list

BrunoGrandePhD:
When a user doesn't specify values for the `required_tests` or `skipped_tests` CLI options, which each allow multiple values, Typer interprets this as an empty list despite the default value being set to `None` ([source](https://github.com/Sage-Bionetworks-Workflows/py-dcqc/blob/eff93a22cdb29f0ef229d9e5ebf07f26dd8eec77/src/dcqc/main.py#L31-L32)). This introduces a bug because the `SuiteABC` class interprets an empty list and `None` differently during instantiation. With `None`, it will follow a default behavior for generating the list. In the case of skipped tests, it remains empty. However, for the case of required tests, the default behavior is currently defined as all tier-1 and tier-2 tests. 

I wish there was a more elegant fix. I tried using a callback function ([docs](https://typer.tiangolo.com/tutorial/options/callback-and-context/)) to automatically convert an empty list to `None`, but this didn't work. I was running into an error because Typer seems to expect an iterable after the callback function is called. I also tried the `Optional[List[str]]` type hint so I didn't have to reassign values to the `_maybe` variables, but this caused another issue. Obviously, Typer is still a young package, so it has some rough edges with its more advanced features. Some of these could be related to Click though. It's hard to tell.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -33,97 +33,84 @@
         },
         "charset-normalizer": {
             "hashes": [
-                "sha256:00d3ffdaafe92a5dc603cb9bd5111aaa36dfa187c8285c543be562e61b755f6b",
-                "sha256:024e606be3ed92216e2b6952ed859d86b4cfa52cd5bc5f050e7dc28f9b43ec42",
-                "sha256:0298eafff88c99982a4cf66ba2efa1128e4ddaca0b05eec4c456bbc7db691d8d",
-                "sha256:02a51034802cbf38db3f89c66fb5d2ec57e6fe7ef2f4a44d070a593c3688667b",
-                "sha256:083c8d17153ecb403e5e1eb76a7ef4babfc2c48d58899c98fcaa04833e7a2f9a",
-                "sha256:0a11e971ed097d24c534c037d298ad32c6ce81a45736d31e0ff0ad37ab437d59",
-                "sha256:0bf2dae5291758b6f84cf923bfaa285632816007db0330002fa1de38bfcb7154",
-                "sha256:0c0a590235ccd933d9892c627dec5bc7511ce6ad6c1011fdf5b11363022746c1",
-                "sha256:0f438ae3532723fb6ead77e7c604be7c8374094ef4ee2c5e03a3a17f1fca256c",
-                "sha256:109487860ef6a328f3eec66f2bf78b0b72400280d8f8ea05f69c51644ba6521a",
-                "sha256:11b53acf2411c3b09e6af37e4b9005cba376c872503c8f28218c7243582df45d",
-                "sha256:12db3b2c533c23ab812c2b25934f60383361f8a376ae272665f8e48b88e8e1c6",
-                "sha256:14e76c0f23218b8f46c4d87018ca2e441535aed3632ca134b10239dfb6dadd6b",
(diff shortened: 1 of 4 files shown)


Pipfile.lock
BrunoGrandePhD, line 1: I tried avoiding this update, but I was running into warnings when I tried recreating a venv from the previous version, so I took the hint that it might be worthwhile updating the lockfile.

src/dcqc/target.py
```diff
@@ -65,10 +64,30 @@ def get_file_type(self) -> FileType:
...
+
+        Raises:
+            ValueError: If the parent directory of the
+                destination does not exist.
+            FileExistsError: If the destination file already
+                exists and ``overwrite`` was not enabled.
+
+        Returns:
+            The path of the local copy.
+        """
```
BrunoGrandePhD, line 90: You can ignore this. I tried to avoid docstring duplication with `wraps()`, but it was causing an error in VS Code because of the parameters. It's unrelated to the PR, but I deemed it small enough to sneak in.

tests/test_main.py
```diff
@@ -113,6 +115,9 @@ def test_create_suite(get_data, get_output):
     check_command_result(result)
     assert output_path.exists()
 
+    suite = JsonParser.parse_object(output_path, SuiteABC)
+    assert len(suite.required_tests) > 0
```
BrunoGrandePhD, line 119: This assertion fails with the original version but passes with the new version. I want to make it a habit to add tests whenever bugs surface to prevent regression.

thomasyu888 (approved):
🔥  Excellent!  What I did was I ran through a series of commands but didn't specify any required tests:

```
dcqc create-targets tests/data/files.csv targets -f
dcqc stage-target targets/target-0007.json tsv_target.json staged_targets -f
dcqc create-tests targets/target-0007.json test_files -f
dcqc compute-test test_files/target-0007.Md5ChecksumTest.json test_results/md5_result.json -f
dcqc compute-test test_files/target-0007.FileExtensionTest.json test_results/file_ext_result.json -f
dcqc create-suite tsv_suite.json test_results/file_ext_result.json test_results/md5_result.json -f
```

Nit: I didn't notice this the first time around, but let's avoid double letter abbreviations in the cli.  I'll open a PR to patch that up: https://clig.dev/#arguments-and-flags

thomasyu888 (approved):

src/dcqc/main.py
```diff
@@ -92,8 +92,12 @@ def create_tests(
     """Create test JSON files from a target JSON file"""
     output_dir.mkdir(parents=True, exist_ok=True)
 
+    # Interpret empty lists from CLI as None (to auto-generate values)
+    required_tests_maybe = required_tests if len(required_tests) > 0 else None
```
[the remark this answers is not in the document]
BrunoGrandePhD: I addressed this here: https://github.com/Sage-Bionetworks-Workflows/py-dcqc/pull/25/commits/f61d01571e16a46e84469392d450763c9e32d8f4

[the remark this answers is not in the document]
BrunoGrandePhD: I'll fix this in https://github.com/Sage-Bionetworks-Workflows/py-dcqc/pull/25

PR #25, opened 2023-03-31
[ORCA-167] Stage files lazily

BrunoGrandePhD:
This PR removes the `stage-files` CLI command in favor of staging the files when they are needed. For internal tests, this happens during the test itself. For external tests, this happens when preparing the process to be run. This is enabled by the `get_files()` method, which retrieves the files while staging them.

### DIFF
--- a/src/dcqc/main.py
+++ b/src/dcqc/main.py
@@ -1,7 +1,7 @@
 import sys
 from csv import DictWriter
 from pathlib import Path
-from typing import List, Optional
+from typing import List
 
 from typer import Argument, Exit, Option, Typer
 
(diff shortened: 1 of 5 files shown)


src/dcqc/file.py
```diff
@@ -334,7 +334,9 @@ def stage(
             if self._local_path is not None:
                 return self._local_path
             else:
-                destination_str = mkdtemp()
+                # TODO: This prefix is used by nf-dcqc to easily find the staged file.
```
thomasyu888, line 337: +1 on this TODO:.  Will file a jira ticket for this:

                # TODO: This prefix is used by nf-dcqc to easily find the staged file.
                #       It might be worth using a DCQCTMPDIR to avoid hard-coding this.
                destination_str = mkdtemp(prefix="dcqc-staged-")

thomasyu888 (approved):
💯 .  Always good when we can remove a cli.

```
dcqc create-targets tests/data/files.csv targets -f
# dcqc stage-target targets/target-0007.json tsv_target.json staged_targets -f
dcqc create-tests targets/target-0007.json test_files -f
dcqc compute-test test_files/target-0007.Md5ChecksumTest.json test_results/md5_result.json -f
dcqc compute-test test_files/target-0007.FileExtensionTest.json test_results/file_ext_result.json -f
dcqc create-suite tsv_suite.json test_results/file_ext_result.json test_results/md5_result.json -f

cat tsv_suite.json
```

I confirm this is installable and works!

PR #26, opened 2023-04-05
[ORCA-136] Add rudimental PHI detection by doing a grep

thomasyu888:
We might come up with a more refined contribution method, but for now, we have pulled your changes into a feature branch @adamjtaylor .  Thanks again for your contribution!

The changes here add the `GrepDateTest` to the TIFF file suite which extends to the `OME-TIFF` suite.  There are also tests that are paired with this. 

One reason this was pulled into a feature branch was that github secrets aren't accessible via forked contributions.

Edit: I modified some tests because we added to TIFF tests.


src/dcqc/suites/suites.py
```diff
@@ -22,7 +22,7 @@ class JsonLdSuite(JsonSuite):
 
 class TiffSuite(FileSuite):
     file_type = FileType.get_file_type("TIFF")
-    add_tests = (tests.LibTiffInfoTest,)
+    add_tests = (tests.LibTiffInfoTest, tests.GrepDateTest)
```
BrunoGrandePhD, line 25: Just to confirm, there's a non-binary portion of TIFF files that can be grep'ed? If the intent is to scan the metadata portion of the TIFF file, then we can revisit this approach once we add support for scanning a portion of a remote file.

tests/test_suites.py
```diff
@@ -4,15 +4,18 @@
...
 FileType("None", ())
 FileType("Unpaired", ())
 
 
 class RedundantFileSuite(TiffSuite):
     file_type = FileType.get_file_type("None")
-    del_tests = (LibTiffInfoTest,)
+    del_tests = (
+        LibTiffInfoTest,
+        GrepDateTest,
```
BrunoGrandePhD, line 17: Note to self: It might be nice to be able to specify strings here so you don't have to import every test that you're adding or removing.

BrunoGrandePhD (approved):

src/dcqc/suites/suites.py
thomasyu888: Yes, @adamjtaylor confirmed that this works

tests/test_suites.py
thomasyu888: Will create a ticket to track this.

src/dcqc/tests/tests.py
```diff
@@ -136,3 +136,24 @@ def generate_process(self) -> Process:
             command_args=command_args,
         )
         return process
+
+
+class GrepDateTest(ExternalTestMixin, TestABC):
+    tier = 2
```
thomasyu888, line 142:
```suggestion
    tier = 4
```

PR #30, opened 2023-04-13
Fix GrepDateTest

adamjtaylor:
Amends `GrepDateTest` to
- Negate the exit code such that a match returns exit 1 (test fails) and no match returns exit 0 (test passes)
- Removes `-w` option so that terms such as `<AcquisitionDate>` where the risk term is not a isolated word now match
- Adds `-a` option to explicitly treat binary files as text
- Adds `-q` option to suppress output as is not needed by the test

Note that for TIFF files this only searches within values of Tags where the content is a string, such as `ImageDescription`
We may wish to add an additional test that searches the `DateTime` tag itself.

### DIFF
--- a/src/dcqc/tests/tests.py
+++ b/src/dcqc/tests/tests.py
@@ -145,11 +145,13 @@ class GrepDateTest(ExternalTestMixin, TestABC):
         file = self.get_file()
         path = file.local_path.as_posix()
         command_args = [
+            "!"            # negate exit status
             "grep",
-            "-E",
-            "-w",
-            "-i",
-            "'date|time'",
+            "-E",          # extended regular expression
+            "-i",          # case insensitive
+            "-a",          # treat input as text
+            "-q",          # suppress output
+            "'date|time'", # match date or time
             path,
         ]
         process = Process(


adamjtaylor:
Tests failing as this is a branch on my fork.
@thomasyu888 could I have my permissions in this repo elevated so that I can create branches in the repo and assign reviewers.

adamjtaylor:
As this test is now part of the TiffSuite, I am tempted to extend it here to grep within the output of `tifftools dump`

eg 
`! tifftools dump  image.tiff | grep -Eiaq 'date|time' `

This would allow for the names of tags such as `DateTime` to be matched within the same test rather than having to run a separate test.

The alternative is to add a new test, `TiffTag` test. I would probablly still use a grep on `tifftools dump` to do this to ensure all IFDs and subIFDs are looked at.

eg.
`! tifftools dump  image.tif | grep -aq 'DateTime 306 (0x132) ASCII'`

thomasyu888:
Thanks @adamjtaylor I'm happy to give you write permission to the repo.  One second while I make some adjustments.  I like what you've proposed for a test.  We can certainly create a new test OR just repurpose this `GrepDateTest` to be `TiffDateTimeTagTest`

PR #31, opened 2023-04-13
Fix grep date test

thomasyu888:
Amends `GrepDateTest` to
- Negate the exit code such that a match returns exit 1 (test fails) and no match returns exit 0 (test passes)
- Removes `-w` option so that terms such as `<AcquisitionDate>` where the risk term is not a isolated word now match
- Adds `-a` option to explicitly treat binary files as text
- Adds `-q` option to suppress output as is not needed by the test

Note that for TIFF files this only searches within values of Tags where the content is a string, such as `ImageDescription`
We may wish to add an additional test that searches the `DateTime` tag itself.

### DIFF
--- a/src/dcqc/tests/tests.py
+++ b/src/dcqc/tests/tests.py
@@ -145,13 +145,12 @@ class GrepDateTest(ExternalTestMixin, TestABC):
         file = self.get_file()
         path = file.local_path.as_posix()
         command_args = [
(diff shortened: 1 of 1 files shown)


thomasyu888:
More discussions here: https://github.com/Sage-Bionetworks-Workflows/py-dcqc/pull/30#issuecomment-1506766484

PR #35, opened 2023-06-02
[ORCA-231] Fix External Tests

BWMac:
This PR is a WIP.

This PR makes the following changes to ensure that external tests work the same locally and on Tower, and to improve the overall functioning of external tests in DCQC.

1. Implementation of the `pass_code` class attribute for external tests.

All external test classes now have the `pass_code` attribute which will either be `"0"` or `"1"` depending on the desired behavior for a test to have passed. For example, a passing scenario for the `TiffInfoTest` is that the `tiffinfo` command is able to be executed successfully and therefore the `pass_code` should be `"0"`. On the other hand, in the case of the `GrepDateTest`, the test is considered to pass when the `grep` command fails so its `pass_code` is `"1"`. To use this new attribute in the evaluation of files in DCQC, I have updated the `_interpret_process_outputs` function within the `ExternalTestMixin` class to use it rather than always expect a code of `"0"` for passing tests.

By implementing this change, it is no longer necessary to negate exit statuses in the command arguments for tests and we are not relying on the implicit meaning of exit codes.

2. Addition of `short_string_path` function to `ExternalTestMixin`.

To solve the file-mounting problems we are facing for `nf-dcqc` runs on Tower in the short-term, I have implemented this function which truncates the path to a staged to the format `dcqc-staged-x/file`. This method ensures that the pipeline behaves the same locally as it does on Tower and that the mounted file is able to be accessed in the `RUN_PROCESS` step. 

Conversations with Seqera are ongoing about the best way to handle our use case, and future work will likely involve improvements to `nf-dcqc` which will make this change unnecessary.

3. All affected tests have been updated to reflect the abovementioned changes.


src/dcqc/tests/base_test.py
```diff
@@ -195,12 +196,22 @@ def _interpret_process_outputs(self, outputs: dict[str, Path]) -> TestStatus:
...
         exit_code = exit_code.strip()
-        if exit_code == "0":
+        if exit_code == self.pass_code:
             status = TestStatus.PASS
         else:
             status = TestStatus.FAIL
         return status
 
+    # TODO: make changes to this package or the nextflow workflow so that file mounting is handled cleaner
+    def short_string_path(self, path: Path, substring: str) -> str:
```
thomasyu888, line 206: Self isn't used, so can this be a staticmethod? Can we also make this a "hidden/helper" function with _ in the beginning?

Alternatively, we could put this in a utility module but I see why this is here.

src/dcqc/tests/libtiff_info_test.py
```diff
@@ -4,13 +4,16 @@
 
 class LibTiffInfoTest(ExternalBaseTest):
     tier = 2
+    pass_code = "0"
     target: SingleTarget
 
     def generate_process(self) -> Process:
         path = self.target.file.stage()
+        string_path = self.short_string_path(path, "dcqc-staged-")
```
thomasyu888, line 12: Right now, I think "dcqc-staged" is hardcoded in the package, I know there's a TODO to include this as a constant - I'll file a Jira ticket for it.

src/dcqc/tests/base_test.py
BWMac: changed to a staticmethod starting with "_" in the latest commit.

I considered putting this in a utility module, but I figured that while it is only used by external tests it should stay here.

src/dcqc/tests/base_test.py
```diff
@@ -195,12 +196,24 @@ def _interpret_process_outputs(self, outputs: dict[str, Path]) -> TestStatus:
...
         else:
             status = TestStatus.FAIL
         return status
 
+    # TODO: make changes to this package or the nextflow
+    # workflow so that file mounting is handled cleaner
+    @staticmethod
+    def _short_string_path(path: Path, substring: str) -> str:
+        # get index where staged folder is
+        index = [i for i, item in enumerate(path.parts) if substring in item][0]
```
thomasyu888, line 210: Nit: I don't foresee this happening, but what happens if the substring doesn't exist? (Since we are the ones always refining the substring)

tests/test_tests.py
```diff
@@ -183,23 +184,24 @@ def test_that_a_process_can_be_serialized_and_deserialized():
...
     tiff_file = test_files["tiff"]
     target = SingleTarget(tiff_file)
     with TemporaryDirectory() as tmp_dir:
         path_0 = Path(tmp_dir, "code_0.txt")
         path_1 = Path(tmp_dir, "code_1.txt")
         path_0.write_text("0")
         path_1.write_text("1")
-        good_outputs = {"std_out": path_1, "std_err": path_1, "exit_code": path_0}
-        bad_outputs = {"std_out": path_0, "std_err": path_0, "exit_code": path_1}
+        zero_outputs = {"std_out": path_1, "std_err": path_1, "exit_code": path_0}
```
thomasyu888, line 195: Just an observation, but it took me a second to realize zero outputs here relates to the exit code and not the number of outputs.

src/dcqc/tests/base_test.py
```diff
@@ -195,12 +196,24 @@ def _interpret_process_outputs(self, outputs: dict[str, Path]) -> TestStatus:
...
+        if exit_code == self.pass_code:
             status = TestStatus.PASS
         else:
             status = TestStatus.FAIL
         return status
 
+    # TODO: make changes to this package or the nextflow
+    # workflow so that file mounting is handled cleaner
+    @staticmethod
+    def _short_string_path(path: Path, substring: str) -> str:
```
thomasyu888, line 208: Is there a test for this function?

src/dcqc/tests/base_test.py
BWMac: if the substring is not present in the path, it will error out on line 212 because the `index` will not exist.

BWMac: I updated the logic to make it a little cleaner (stopped using the 0 index logic) and to handle the case where the substring is not present in the path. Also added a test for that error raising in that case.

src/dcqc/tests/base_test.py
BWMac: my bad, added in latest commit

tests/test_tests.py
BWMac: I reverted back to the old `pass`- and `fail` outputs variable naming. I think the comment in each test is enough for comprehension, and I agree that this change was confusing.

src/dcqc/tests/base_test.py
```diff
@@ -195,12 +196,27 @@ def _interpret_process_outputs(self, outputs: dict[str, Path]) -> TestStatus:
...
+        # chech if substring is in path
+        if substring not in path.as_posix():
+            raise ValueError(f"{substring} not in {path}")
+        # get index where staged folder is
+        index = next(i for i, item in enumerate(path.parts) if substring in item)
```
thomasyu888, line 213: In the unlikely scenario there is something like:

`substring-A/substring-B/substring-C/your_file.txt`

That takes the first occurrence right? I think I'm ok with this. If the letters dcqc-staged ever gets arranged randomly in that order by nextflow, I'll just cry a little bit.

thomasyu888 (approved):
🔥 awesome! I'm going to pre-approve, but has this been tested both locally and on tower?

src/dcqc/tests/base_test.py
BWMac: Yes, this finds the first occurrence and then chops off anything that came before the substring. I think that is fine though because the only purpose of this string manipulation is to get us past the Nextflow-generated piece of the path. 

But yes, I suppose it is possible for Nextflow to by pure chance create a path with the substring in it. The way we are using it currently we are saved from this because we have a special character "-" in our substring which I don't think Nextflow would use in its generated paths.

thomasyu888:
🔥 nice work here! Im excited about this update.

thomasyu888 (approved):

PR #37, opened 2023-06-07
[ORCA-125] Implement `PairedFastqParityTest` as our first multi-file QC test

BrunoGrandePhD:
This PR builds on top of the recent refactoring with tests and targets and implements our first multi-file QC test. `PairedFastqParityTest` is a simple test that confirms whether paired FASTQ files have the same number of rows. While I added code tests and associated test data to ensure that the QC test works as expected, it's currently not going to be picked up by the CSV parser that generates the tests based on the list of input files. I decided that should be tackled separately to keep our PRs focused and manageable.


src/dcqc/target.py
```diff
@@ -128,3 +121,18 @@ def get_file_type(self) -> FileType:
             The file type object.
         """
         return self.file.get_file_type()
+
+
+@dataclass(init=False)
+class PairedTarget(BaseTarget):
```
thomasyu888, line 127: Discussion: Would it be more flexible to name this `MultiFileTarget`?

src/dcqc/tests/paired_fastq_parity_test.py
```diff
@@ -0,0 +1,57 @@
...
+        """Open a FASTQ file regardless of compression.
+
+        Args:
+            path: Path to the FASTQ file.
+
+        Returns:
+            Opened FASTQ file (in text mode).
+        """
+        if path.name.endswith(".gz"):
+            return gzip.open(path, "rt")
```
thomasyu888, line 55: This is two tests combined in one. 

If the file ends in .gz but isn't actually gunzipped, this will fail causing a failure code.

thomasyu888 (approved):
🔥 Excellent!  Going to pre-approve, but just some discussion points.

src/dcqc/tests/paired_fastq_parity_test.py
BrunoGrandePhD: You're right. Ideally, this logic would live in the `File` class. I'll leave a TODO comment about this.

PR #38, opened 2023-06-09
adds SuiteStatus and applies changes as needed

BWMac:


src/dcqc/suites/suite_abc.py
```diff
@@ -207,24 +216,27 @@ def compute_tests(self) -> None:
...
+                    self._status = SuiteStatus.RED
+                    break
+            else:
+                if test_status == TestStatus.FAIL:
+                    self._status = SuiteStatus.AMBER
         return self._status
 
     def to_dict(self) -> SerializedObject:
         suite_status = self.compute_status()
+        print(suite_status)
```
thomasyu888, line 239: Nit: do we need this print?

thomasyu888 (approved):
🔥 great work! Just a minor nit - I haven't had the time to install and run a test command but I trust that it works.

src/dcqc/suites/suite_abc.py
BWMac: Nope! I accidentally left that in from my experimenting. removed in the latest commit.

tests/data/suite.json
```diff
@@ -22,7 +22,7 @@
     "skipped_tests": [
       "LibTiffInfoTest"
     ],
-    "status": "passed"
+    "status": "GREEN"
```
BrunoGrandePhD, line 25: Did you update this manually or run the `tests/data/generate.py` script? If the former, I suggest running the script to ensure that the files are generated consistently. It also acts as a bit of a test. 

Actually, it might make sense to convert the code in `generate.py` to Pytest fixtures, which generate the data files dynamically during testing. This way, it's automatic. This would be a good future refactoring task.

src/dcqc/suites/suite_abc.py
```diff
@@ -207,20 +216,22 @@ def compute_tests(self) -> None:
...
-            if test_name not in self.required_tests:
-                continue
             test_status = test.get_status()
-            self._status = test_status
-            if self._status == TestStatus.FAIL:
-                break
+            if test_name in self.required_tests:
+                if test_status == TestStatus.FAIL:
+                    self._status = SuiteStatus.RED
+                    break
```
BrunoGrandePhD, line 231: Nit: You could also return `self._status` here.

BrunoGrandePhD (approved):
A perfect PR. Small, self-contained, and focused on one change. 🚀

tests/data/suite.json
BWMac: I did this manually originally. The latest commit adds minor changes from running `generate.py` and adds a TODO to move that logic to pytest fictures.

src/dcqc/suites/suite_abc.py
BWMac: changed in the latest commit

PR #39, opened 2023-06-09
[ORCA-175] Add `CsvUpdater` and `update_csv` command

BWMac:
This PR follows on from #38 and implements a `CsvUpdater` class (and needed tests) which will be used to update the input CSV file for the DCQC run with a new column indicating the `SuiteStatus` for each file following processing. It also adds a new CLI command `update_csv` which will apply this new class to make the needed changes. The logic for updating CSV files is agnostic to single/multi-target QC runs, and uses the file URL as an ID to evaluate all suites generated against each individual target file. Necessary example files `input.csv`, `output.csv`, and `suites.json` are also added to `tests/data`. 

`suite_abc.py` is also updated to include a `get_status` method for accessing the `SuiteStatus` from outside of the class. 

edit: I am marking this as ready for review after successfully testing these changes in `nf-dcqc`.


src/dcqc/suites/suite_abc.py
```diff
@@ -292,3 +292,9 @@ def from_dict(cls, dictionary: SerializedObject) -> SuiteABC:
     def get_base_class(cls):
         """Retrieve base class."""
         return SuiteABC
+
+    def get_status(self, compute_ok: bool = True) -> SuiteStatus:
```
thomasyu888, line 296: Maybe I missed this, but when would you use `compute_ok`?

src/dcqc/main.py
```diff
@@ -204,3 +205,15 @@ def qc_file(
     report = JsonReport()
     suite_json = report.generate(suite)
     json.dump(suite_json, sys.stdout, indent=2)
+
+
+@app.command()
+def update_csv(
```
thomasyu888, line 211: This is a great addition! Can you spend a tiny amount of time looking into how to write to nextflow tower reports? (That can be a separate ticket, but if it's super easy, maybe it could be bundled)

src/dcqc/updaters.py
```diff
@@ -0,0 +1,56 @@
+from csv import DictWriter
+from pathlib import Path
+from typing import Dict, List
+
+from dcqc.parsers import CsvParser
+from dcqc.suites.suite_abc import SuiteABC
+
+
+class CsvUpdater:
```
thomasyu888, line 9: Can you add dataclass decorator here? Or is there a reason you left it out?

src/dcqc/updaters.py
```diff
@@ -0,0 +1,56 @@
...
+    input_path: Path
+    output_path: Path
+    parser: CsvParser
+
+    def __init__(self, input_path: Path, output_path: Path):
+        self.output_path = output_path
+        self.parser = CsvParser(input_path)
```
thomasyu888, line 16: If you don't need to specify the parser, this could be in a post init.

src/dcqc/updaters.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def update(self, suites: List[SuiteABC]):
+        suite_dict: Dict[
```
thomasyu888, line 19: What was the error that mypy gave ?

Look into collections.defaultdict.  that might be helpful and will simplify your logic below.

src/dcqc/updaters.py
```diff
@@ -0,0 +1,56 @@
...
+            str, List[str]
+        ] = {}  # mypy made me do this, but only here. not sure why
+        # {url: [list_of_statuses]} data structure to allow for multi-file targets
+        for suite in suites:
```
thomasyu888, line 23: So with multi file targets, the current logic is that a particular file will have multiple suites.

It doesn't look like you're looping through the multiple files within a suite though (not sure you need to, just an observation)

src/dcqc/updaters.py
```diff
@@ -0,0 +1,56 @@
...
+            elif "AMBER" in statuses:
+                collapsed_dict[url] = "AMBER"
+            elif "GREEN" in statuses:
+                collapsed_dict[url] = "GREEN"
+            else:
+                collapsed_dict[url] = "NONE"
+        # Create CSV data structure
+        row_list = []
+        for row in self.parser.list_rows():
+            csv_data = row[1]
```
thomasyu888, line 44: Remind me, does list_rows return a tuple?

src/dcqc/updaters.py
```diff
@@ -0,0 +1,56 @@
...
+            csv_data["dcqc_status"] = collapsed_dict[csv_data["url"]]
+            row_list.append(csv_data)
+
+        keys = row_list[0].keys()
```
thomasyu888, line 48: Not saying that this happens, but what does the code do if we pass in an empty manifest?

thomasyu888 (commented):
Nice! This is my initial feedback - will review more tomorrow.

src/dcqc/suites/suite_abc.py
BWMac: this was copied over from the `get_status` method in the tests module and I forgot to remove `compute_ok`. removed in latest commit.

src/dcqc/updaters.py
BWMac: it returns a generator object which allows you to loop through row tuples that look like `(index, dict)`

src/dcqc/updaters.py
BWMac: it will throw an `IndexError` on line 51. should I catch that error, or try to rework this so it is not an issue?

src/dcqc/updaters.py
BWMac: the error was `Need type annotation for "suite_dict" (hint: "suite_dict: Dict[<type>, <type>] = ...")  [var-annotated]`. 

I will look into defaultdict

BWMac: I added defaultdict to the logic. The error is gone and the logic is simpler. Thanks!

src/dcqc/updaters.py
BWMac: So does that mean that for a file pair, each member of the pair would have an individual suite and then they would both be part of a paired suite? Meaning that the `url` field would have two URLs in that case? Is there an existing example somewhere that I can use?

src/dcqc/main.py
BWMac: Would we just want to display the output CSV file like [this](https://help.tower.nf/22.1/reports/overview/#visualising-reports)?

src/dcqc/updaters.py
BWMac: I added logic to catch the case where the manifest is empty and therefore nothing is appended to `row_list`

src/dcqc/updaters.py
thomasyu888: Currently, the functionality isn't fully used, we need to add the ability for DCQC to leverage the multi-file feature that was added.  Lets add a TODO for now.

tests/test_suites.py
```diff
@@ -124,3 +126,16 @@ def test_that_a_suite_will_consider_required_tests_when_passing(test_targets):
     suite = SuiteABC.from_target(target, required_tests)
     suite_status = suite.compute_status()
     assert suite_status == SuiteStatus.GREEN
+
+
+def test_that_status_is_computed_if_not_already_assigned(test_targets):
+    with patch.object(
+        SuiteABC, "compute_status", return_value=SuiteStatus.GREEN
```
thomasyu888, line 133: What is your rationale for patching this here?  If some code in `compute_status` is wrong, would that cause the wrong SuiteStatus to be returned?

tests/conftest.py
```diff
@@ -129,3 +130,35 @@ def _get_output(filename: str) -> Path:
...
+    for url, status in mock_dict_single.items():
+        suite = MagicMock(cls=SuiteABC)
+        suite.target.files[0].url = url
+        suite.get_status.return_value = status
+        mocked_suites.append(suite)
+    return mocked_suites
+
+
+@pytest.fixture
+def mocked_suites_multi_targets():
```
thomasyu888, line 152: Can we comment this out until its ready?

tests/test_suites.py
BWMac: I wanted to be able to assert that the function was called, as it should when the status has not yet been computed and also be able to control which status to expect for this test.

src/dcqc/updaters.py
```diff
@@ -0,0 +1,58 @@
...
+        collapsed_dict = {}
+        for url, statuses in suite_dict.items():
+            if "RED" in statuses:
+                collapsed_dict[url] = "RED"
+            elif "AMBER" in statuses:
+                collapsed_dict[url] = "AMBER"
+            elif "GREEN" in statuses:
+                collapsed_dict[url] = "GREEN"
+            else:
+                collapsed_dict[url] = "NONE"
```
thomasyu888, line 39: Do we need to cover this test case? 

When will this be "NONE"?

thomasyu888 (approved):
🔥 @BWMac Nice work!  I'm going to pre-approve, but please add the missing test case for when there is no status.

src/dcqc/updaters.py
BWMac: It would only be None in a case where all of the Suites for a file have not had any statuses assigned. I don't think that it is technically possible for this to be the case the way everything is currently written.

thomasyu888: What about for files that aren't supported in dcqc?

BWMac: unsupported file types just error out [here](https://github.com/Sage-Bionetworks-Workflows/py-dcqc/blob/c7262f647cf584123d5de5b4611d7bbd9127eeac/src/dcqc/file.py#L97)

BWMac: I am concluding that it should not be possible for the final status of a suite to be None, so I will remove this case handling.

src/dcqc/updaters.py
```diff
@@ -33,10 +33,8 @@ def update(self, suites: List[SuiteABC]):
                 collapsed_dict[url] = "RED"
             elif "AMBER" in statuses:
                 collapsed_dict[url] = "AMBER"
-            elif "GREEN" in statuses:
-                collapsed_dict[url] = "GREEN"
             else:
-                collapsed_dict[url] = "NONE"
+                collapsed_dict[url] = "GREEN"
```
thomasyu888, line 37: Would this potentially lead to false positives?

BWMac: I think that could only happen if there were more statuses added

thomasyu888: Does it make sense to keep the None then?

BWMac: Yeah, I guess so. Better safe than sorry. I'll add a test for that case as well

PR #40, opened 2023-06-21
[ORCA-234] Improve file path handling

BWMac:
This is a WIP.

This PR simplifies external tests, and enables the changes in the `nf-dcqc` PR linked below by removing the string manipulation function introduced in #35 and instead passing the name of the staged file wrapped in quotes, thereby emulating the path which Nextflow expects and would use in a command

It is paired with this [PR](https://github.com/Sage-Bionetworks-Workflows/nf-dcqc/pull/7) on `nf-dcqc`.

### DIFF
--- a/src/dcqc/file.py
+++ b/src/dcqc/file.py
@@ -125,7 +125,7 @@ class File(SerializableMixin):
             current work directory (default).
     """
 
-    DCQCTMPDIR: ClassVar[str] = "dcqc-staged-"
+    tmp_dir: ClassVar[str] = "dcqc-staged-"
 
     _serialized_properties = ["name", "local_path"]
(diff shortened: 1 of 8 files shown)


src/dcqc/file.py
```diff
@@ -125,6 +125,8 @@ class File(SerializableMixin):
             current work directory (default).
     """
 
+    DCQCTMPDIR: ClassVar[str] = "dcqc-staged-"
```
thomasyu888, line 128: Nit: since this is a class var now, I don't think it needs to be all caps.

thomasyu888 (approved):
🤖 Awesome fix!  Thanks!

PR #41, opened 2023-06-23
[ORCA-236] Adds missing code handling and tests for external tests

BWMac:
This PR is a bug fix in response to @adamjtaylor reporting that the `BioFormatsTest` was not working.

Both the `BioFormatsTest` and `OmeXmlSchemaTest` were not updated along with the other external tests in #35. They were missed due to missing test coverage for their ability to evaluate exit codes.

Both tests are updated to evaluate exit codes the same way the other external tests do, and the missing test coverage is rectified in this PR.


thomasyu888 (approved):
🔥 Thanks Brad!  Are there other external tests that we need to add this to?

BWMac:
@thomasyu888 No, this brings all 5 external tests up to date

PR #43, opened 2023-07-20
adds double dile staging prevention

BWMac:


src/dcqc/file.py
```diff
@@ -308,6 +309,27 @@ def is_file_local(self) -> bool:
         """
         return self._local_path is not None
 
+    def already_staged(self) -> list[Path]:
```
thomasyu888, line 312: Just a high level question, but do you think this is something that should be tackled in fs-synapse?

Basically if a file has been staged, should fs-synapse know to use the existing staged file?

src/dcqc/file.py
```diff
@@ -308,6 +309,27 @@ def is_file_local(self) -> bool:
...
+
+        Returns:
+            staged_file_paths (list): List of already staged file paths.
+            Empty list if file has not been staged.
+
+        Raises:
+            FileExistsError: If the file has already been staged more than once.
+            This would cause a name collision in Nextflow.
+        """
+        path_str = os.path.join(gettempdir(), self.tmp_dir + "*", self.name)
```
thomasyu888, line 323: Is the star a risk for files that have the same name but different content?

src/dcqc/file.py
```diff
@@ -308,6 +309,27 @@ def is_file_local(self) -> bool:
...
+        staged_file_strs = glob.glob(path_str)
+        staged_file_paths = [Path(path) for path in staged_file_strs]
+        if len(staged_file_paths) > 1:
```
thomasyu888, line 326: I like this - is this to account for potential staging issues that already exist?

Also nit: you can just do `if staged_file_paths`

src/dcqc/file.py
BWMac: maybe, having a function that just lets you check if it has already been staged in fs-synapse might be useful

src/dcqc/file.py
BWMac: the star subs for the random string that is appended to the `dcqc-staged-` prefix so the path is then `dcqc-staged-*/file.txt`. So it should only get those paths where the same file has been staged in a different parent directory.

src/dcqc/file.py
BWMac: right, this is to catch any weird edge case where a file has already been staged more than once and will eventually cause a name collision in nextflow. I think I need to use `len(staged_file_paths) > 1` because I want to allow for the scenario where there is one member of that list. In that case, that one member is the already staged location which we can then use.

thomasyu888: Ah right, I forgot about that...

src/dcqc/file.py
thomasyu888: Hmmm, i had a closer look and apologies. The "stage" function is specific to this package. Would it make sense to incorporate the logic you've implemented in already_staged in there?

BWMac: I did that originally, but because it's a fair bit of logic and `stage` is already large I figured a separate function made sense.

thomasyu888: Makes sense to me: although this does seem like a helper function.

thomasyu888 (approved):
🔥 I'm going to pre-approve, this looks great. Just some comments.

tests/test_file.py
```diff
@@ -62,6 +64,27 @@ def test_that_a_local_temporary_path_is_created_when_staging_a_remote_file(test_
     assert remote_file.local_path == staged_path
 
 
+def test_that_error_is_raised_when_a_file_has_been_staged_multiple_times(test_files):
```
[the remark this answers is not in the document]
BWMac: Yes, and I also realized that now that we are preventing files from being staged more than once, that outcome is occurring after the first successful execution of `stage` in a test. I.e. we are not testing what we used to be testing. I will add a helper function to the tests that clears out the `dcqc-staged-*` directories after each test calling `stage` so that each one is testing what it is meant to.

PR #47, opened 2023-08-18
[ORCA-268] Adds TiffDateTimeTest

BWMac:
Recently, @adamjtaylor reported that the `GrepDateTest` was not behaving as intended, resulting in TIFF files failing QC that should have passed.

After a brief investigation, I determined that the `GrepDateTest` inexplicably returns inconsistent results for TIFF files when running on Tower. In order to provide the intended functionality of that test but with greater specificity, I have implemented the `TiffDateTimeTest ` using the command suggested by Adam in [slack](https://sagebionetworks.slack.com/archives/C04TA90AP5W/p1690894696353589?thread_ts=1690891617.578949&cid=C04TA90AP5W). This new test dumps the tags attached to a particular TIFF file. It then scans those tags for the substrings `date` and `time` case-insensitively.

Tests in this repo are also updated to use this new test as an example instead of `GrepDateTest`, and tests are added to cover `TiffDateTimeTest`.

This change has been tested both locally and on [Tower](https://tower-dev.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/example-dev-project/watch/1x2jRO9DulS294) successfully.


thomasyu888 (approved):
💯 nice!  Sorry missed this PR!

PR #50, opened 2023-10-30
[ORCA-284] External Test Testing & [ORCA-287] OmeXmlSchemaTest Bug fix

BWMac:
[ORCA-284] Problem:

Currently, there is no way to run external tests without using `nf-dcqc`, and therefore there is no way to easily test them to ensure that they are evaluating files as intended.

[ORCA-284] Solution:

Implement Docker-enabled tests that can run the exact commands that `nf-dcqc` currently does via a local or GitHub Actions Docker environment.

Of note in this implementation:
1. `docker` python package was added as a `testing` dependency
2. `BioFormatsInfoTest` was updated to specify the `OMETiff` file format in its CLI command. The test was producing inconsistent results in local testing without this specified. This change has been tested in `nf-dcqc` and yielded the same consistent results as before.
     - Currently, this test is only used in the `OmeTiffSuite` so this should not cause problems unless 
     it is added to other/new suites.
3. `test_tests.py` has been converted into `test_external_tests.py` (I should have just created the new files and deleted the old, the diff is a little weird), and internal test unit tests have been split off into `test_internal_tests.py`.
4. In order to support Docker-enabled testing, a new class `DockerExecutor` was introduced to `test_external_tests.py`. This class takes the docker image string and command from the `Process` object generated for external tests, in addition to the path to a testing file, and creates a `DockerExecutor` which can run external test commands by invoking its `execute` method.
     - A couple of tweaks were needed to standardize command execution across containers: 
          - The `format_command_for_sh` method escapes special characters in the 
          command and prefaces it with `sh -c` to ensure that commands are executed in a similar 
          fashion across Docker containers
          - `working_dir` was set to `"/"` in `client.containers.run` to ensure that files are able to be 
          accessed once mounted, regardless of `WORKDIR` configurations in specific images.
5. While running these new tests in GitHub actions, they worked in the `ubuntu` runner immediately, but `macos` proved trickier. I did some research and long story short GitHub did add support to easily run Docker in the `macos` runner in this [PR](https://github.com/actions/runner-images/issues/2150), but it has been reported recently that this has stopped working due to changes in dependencies. As an interim solution, I implemented a simple wrapper that causes Docker-enabled tests to be skipped when they are run in a non-Linux environment, rather than adding more complex changes to our CI workflow.

[ORCA-287] Problem:

`OmeXmlSchemaTest` always returns an exit code of 0, even when a file fails.

[ORCA-287] Solution:

The results of the `bftools/xmlvalid` command are actually found in its `stdout`, so instead of evaluating directly on the first `exitcode`, we pipe the output into a `grep` command and look for the substring that indicates success: `"No validation errors found."`. This way, the `pass_code` for the test can remain `"0"`, but when the file fails the test it can produce the expected `"1"`. This change has been tested in `nf-dcqc` and with local Docker execution.

[ORCA-284]: https://sagebionetworks.jira.com/browse/ORCA-284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ORCA-287]: https://sagebionetworks.jira.com/browse/ORCA-287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/test_external_tests.py
```diff
@@ -1,90 +1,121 @@
...
-    assert test_status == TestStatus.PASS
-
-
-def test_that_a_tiff_file_with_good_extensions_is_passed(test_targets):
-    target = test_targets["tiff"]
-    test = tests.FileExtensionTest(target)
-    assert test.get_status() == TestStatus.PASS
-
+def test_that_all_external_tests_inherit_from_the_mixin_first():
+    tests = BaseTest.list_subclasses()
```
BryanFauble, line 16: Code smell:
`Redefining name 'tests' from outer scope (line 10) PylintW0621:redefined-outer-name`

tests/test_external_tests.py
```diff
@@ -1,90 +1,121 @@
+import os
```
BryanFauble, line 1: (No changed needed, but I wanted to share some resources with you).

This article gives a great overview of Behavior Driven Development (BDD) topics that can be applied to testing: https://pythontest.com/strategy/given-when-then-2/

BDD is also heavily used by automation engineers when writing automated end-to-end testing in a number of frameworks, so it's likely you'll run into this eventually.

I find thinking in a GIVEN-WHEN-THEN structure for my tests gives me clearer focus that I am testing a singular thing, I know what my assumptions are, and I know the actions I am taking.

I started to use this structure in the Synapse Python Client:

1. https://github.com/Sage-Bionetworks/synapsePythonClient/commit/0d1721fffd74231e477d8c15dad6a98e0988df53#diff-5056ccf1a22d1e69ce5aeea47808fc00ebe8173df1b70a9eb6cd02d2d38cad51R505
2. https://github.com/Sage-Bionetworks/synapsePythonClient/commit/c6223b56e2b1963b880af33a8c19c9c585875a2b#diff-eb75db4e0b13e5e4845036998dcf3c8c8c5509fb0560b21f5e7d1d21ec448d7bR356

tests/test_external_tests.py
```diff
@@ -1,90 +1,121 @@
...
+        self.std_err = container.logs(stdout=False, stderr=True).decode("utf-8")
 
 
-def test_that_the_jsonld_load_test_works_on_a_correct_file(test_targets):
-    target = test_targets["jsonld"]
-    test = tests.JsonLdLoadTest(target)
-    test_status = test.get_status()
-    assert test_status == TestStatus.PASS
+def test_that_the_libtiff_info_test_command_is_produced(test_targets):
+    target = test_targets["tiff"]
```
BryanFauble, line 93: nit: I could see these static strings through the test being an enum of the possible valid values you could have: https://docs.python.org/3/library/enum.html

tests/test_external_tests.py
```diff
@@ -112,18 +143,33 @@ def test_that_the_libtiff_info_test_correctly_interprets_exit_code_0_and_1(
...
+    test = tests.BioFormatsInfoTest(target)
     process = test.generate_process()
-    assert "tiffinfo" in process.command
+    assert "showinf" in process.command
 
 
-def test_that_the_bioformats_info_test_command_is_produced(test_targets):
-    target = test_targets["tiff"]
+@docker_enabled_test
+def test_that_the_bioformats_info_test_exit_code_is_0_when_it_should_be(test_files):
```
BryanFauble, line 154: Food for thought (feel free to take it or leave it): 
You did a great job splitting the tests up into the various python scripts - One area that could be improved is to have 1 more level of organization within the script by grouping like tests together. For example:

```
class TestBioformats:
    def test_....
    def test_....

class TestTiffDateTimeTest:
    def test_...
```

Further reading: https://docs.pytest.org/en/7.1.x/getting-started.html#group-multiple-tests-in-a-class

BryanFauble (approved):
Just a few comments I left, otherwise LGTM!

tests/test_external_tests.py
BWMac: addressed in latest commit

tests/test_external_tests.py
BWMac: I agree. I already created a [ticket](https://sagebionetworks.jira.com/browse/ORCA-288) for reworking the test file names because I find them to be confusing. I can explore this option as part of that task.

tests/test_external_tests.py
BWMac: I did consider this and have [used test classes](https://github.com/Sage-Bionetworks/agora-data-tools/blob/32e6f941c114934414baa5ffcf6a34075bca7394/tests/test_process.py#L13) in the past

tests/test_external_tests.py
BWMac: This is great food for thought and I will look to incorporate these concepts in future work!

thomasyu888 (approved):
🔥 LGTM!

PR #51, opened 2023-11-07
Correcting the links in the setup.cfg

BryanFauble:
Problem:
URLs are not linking back to the DCQC github - they're linking out to pyscaffold

Solution:
Update URLs to this GH project


thomasyu888 (approved):
🔥 thanks for fixing those!

PR #52, opened 2023-12-12
[ORCA-288] Test Object Naming Refactor

BWMac:
**Problem:**

The test file, target, and suite objects defined in `conftest.py` that we use in our unit tests have confusing names, making it difficult to know what is happening in our tests.

**Solution:**

Change the names to a loose convention of `good_<type_of_file>` for "good" files and `<reason_why_this_file_is_bad_<type_of_file>` for "bad" files.

Notes:

- All tests run and pass as expected after these changes.
- I also went ahead and made a couple of changes to simplify some of the logic in `conftest.py`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,6 +74,8 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python }}
+      - name: Install setuptools
+        run: sudo -H pip install setuptools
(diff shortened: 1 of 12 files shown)


BryanFauble (approved):
LGTM! Thanks for making these tests more clear!

thomasyu888 (approved):
🔥 Looks very straightforward to me.  Unfortunately, tox is somewhat of a headache which is making me wonder if it's worth just sticking to basics.

PR #53, opened 2023-12-13
[ORCA-294] Fix `tox` implementation in CI pipeline

BWMac:
**Problem:**

Our CI pipeline is failing to test our code with the intended Python versions.

**Solution:**

Update `ci.yml` to use the `tox-gh` GH Actions plugin.

Notes:

- Once we were testing the Python versions we intended, it became clear that this codebase was not actually compatible with Python 3.8. I am changing the compatibility to 3.9-3.11 as a result. I do not expect this to impact any users.
- `tox` is no longer a required dependency of the repository because we are installing `tox-gh` during our CI runs according to the recommended [usage](https://github.com/tox-dev/tox-gh#usage).
- This PR will serve as a model for any other repositories we own that use `tox` in their CI pipeline.
- A GH Action run with these changes can be viewed [here](https://github.com/Sage-Bionetworks-Workflows/py-dcqc/actions/runs/7199308971).
     -  Please verify that the versions listed in the logs for the `Run tests` steps under `tox:py` match those expected before approving as a sanity check.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
     strategy:
       matrix:
         python:
-        - "3.8"
+        # - "3.8"
         - "3.11"  # newest Python that is stable
         platform:
         - ubuntu-latest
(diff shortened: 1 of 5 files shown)


thomasyu888 (approved):
🔥 LGTM!

PR #54, opened 2023-12-14
[ORCA-289] Refactor External and Internal Testing Unit Tests as Test Classes

BWMac:
**Problem:**

Unit tests for `py-dcqc` Internal and External tests are disorganized, making it more difficult than it needs to be to contribute new DCQC tests.

**Solution:**

Organize all unit tests that specifically evaluate the functionality of DCQC Internal and External tests into test classes to make it clear what unit tests belong together. It should now be simpler for contributors to add new DCQC tests and any needed unit tests.


thomasyu888 (approved):
🔥 Fantastic, this makes the tests more readable and modular.

PR #55, opened 2024-02-05
Add support for QPTIFF

adamjtaylor:
[QPTIFF](https://docs.openmicroscopy.org/bio-formats/5.9.2/formats/perkinelmer-vectra-qptiff.html) is a TIFF-based image format used by PerkinsElmer Vecra mIF and Akoya Phenocycler CODEX instruments. This PR adds basic support for them by including files with the `.qptiff` extension in the TIFF file type.

### DIFF
--- a/src/dcqc/file.py
+++ b/src/dcqc/file.py
@@ -106,7 +106,7 @@ FileType("*", (), "format_1915")  # To represent all file types
 FileType("TXT", (".txt",), "format_1964")
 FileType("JSON", (".json",), "format_3464")
 FileType("JSON-LD", (".jsonld",), "format_3749")
-FileType("TIFF", (".tif", ".tiff", ".svs", ".scn"), "format_3591")
+FileType("TIFF", (".tif", ".tiff", ".svs", ".scn", "qptiff"), "format_3591")
 FileType("OME-TIFF", (".ome.tif", ".ome.tiff"), "format_3727")
 FileType("TSV", (".tsv"), "format_3475")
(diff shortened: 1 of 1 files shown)


src/dcqc/file.py
```diff
@@ -106,7 +106,7 @@ def get_file_type(cls, file_type: str) -> FileType:
 FileType("TXT", (".txt",), "format_1964")
 FileType("JSON", (".json",), "format_3464")
 FileType("JSON-LD", (".jsonld",), "format_3749")
-FileType("TIFF", (".tif", ".tiff", ".svs", ".scn"), "format_3591")
+FileType("TIFF", (".tif", ".tiff", ".svs", ".scn", "qptiff"), "format_3591")
```
BWMac, line 109:
```suggestion
FileType("TIFF", (".tif", ".tiff", ".svs", ".scn", ".qptiff"), "format_3591")
```

BWMac (commented):
Out of curiosity, have you run any `.qptiff` files through the `TiffSuite`? I'm assuming they would pass all of the normal `.tif` tests (except for file extension until after this is merged), but it might be good to make sure.

BWMac (approved):
LGTM! I'm guessing the one acceptance test is failing because this contribution is coming from a fork which doesn't have a Synapse Auth token with the necessary permissions. The test suite passes locally for me and the change is tiny so I think we are safe to merge this. I will be sure tests all pass before releasing a patch with this change.

Thanks for the contribution @adamjtaylor!

PR #56, opened 2024-05-03
[ORCA-253] Return Failure Reasons to `suites.json` and Failed Tests to `output.csv`

BWMac:
**Problem:**

Currently, the results provided by DCQC to users in the output file are limited to the `dcqc_status` for each file in the original manifest. Additionally, there is no status indicating that a test was unable to be performed on a file due to an error encountered during the execution of an external test.

**Scope:**

This PR adds functionality to capture failure reasons for both internal and external tests, but errors and error reasons for external tests only. I also encountered a complication with external tests where many of them will return the same `exit_code` whether they have failed or errored out. This will be addressed in future work by adding some more complex/specific logic to assess a "true" failure vs. an error. For now, these errors show up as failures, and the `error_reason` is captured instead as `failure_reason`.

**Solution:**

To surface more information about DCQC runs to users, the following changes were made:
- Implemented logic for capturing failure reasons for both external and internal tests. 
     - For external tests, this involved updates to the `ExternalTestMixin` and each individual external test class to capture information about how each test should be interpreted. 
     - For internal tests, this involved adding failure reason strings at each place in the test logic where a test failure would be determined. 
     - As mentioned above, errors and error reasons were also captured for external tests only. 
     - Implemented logic to surface which tests were required, skipped, failed, and errored for each file in the final DCQC output file.

**Testing:**
- All necessary tests and test files were updated in the repository to accommodate these changes.
- New testing conditions were added for capturing the internal test failure reasons where needed.
- The changes above were tested on Nextflow Tower.
     - This [run](https://tower-dev.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/example-dev-project/watch/2cAzi4UDkD3A1b) was used to test a set of files to determine if the overall logic was working.
     - This [run](https://tower-dev.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/example-dev-project/watch/tT3875QOcLsnd) was used to test the suite error  by intentionally providing a bad file path to the `TiffTag306DateTimeTest`
     - Both test runs used a new `test_full` [configuration](https://github.com/Sage-Bionetworks-Workflows/nf-dcqc/pull/11) that I created to test all file types that we support. The files used are housed [here](https://www.synapse.org/#!Synapse:syn59094786) on Synapse.

**Notes:**
- Bumped `fs-synapse` to 2.0 (`synapseclient` version 4.1.1).
- Updated `CONTRIBUTING.md`.
- Added DPE as code owners.


src/dcqc/tests/base_test.py
```diff
@@ -190,16 +206,21 @@ def _find_process_outputs(
...
-        exit_code = outputs["exit_code"].read_text()
-        exit_code = exit_code.strip()
+        exit_code = int(outputs["exit_code"].read_text().strip())
+
         if exit_code == self.pass_code:
             status = TestStatus.PASS
-        else:
+        elif exit_code == self.fail_code:
             status = TestStatus.FAIL
+            self.failure_reason = outputs[self.failure_reason_location].read_text()
```
BryanFauble, line 220: nit: A slightly cleaner way to do:

```
self.failure_reason
and
self.error_reason
```

might be to have just a generic:
```
self.status_reason
```

It is a slight shift in responsibility but I'm thinking about "What if I have a `warning` and I need to return back. Would I need to create a new field named `warning_reason`?".

If you have a generic `status_reason` it'll be "Why is the status the value it is".

CONTRIBUTING.md
```diff
@@ -186,34 +186,78 @@ This often provides additional considerations and avoids unnecessary work.
...
+When contributing an internal test be sure to do the following:
+
+1. Follow the steps above to set up `py-dcqc` and create your contribution.
+
+1. Include a class docstring that describes the purpose of the test.
+
+1. Include the following class attributes:
+
+   - `tier`: An `int` value describing the complexity of the test contributed. `tier` values include:
+     - `1`: File Integrity
```
BryanFauble, line 202: nit: It might be clearer within the test if these are an enum (That could map to an integer). That way instead of just seeing:
```
    tier = 1
```

You will see:
```
    tier = SUBJECTIVE_CONFORMANCE
```

Then when evaluating it in the base class check the enum or int value of the attribute. The tier type would be:
```
tier: Union[int, MyTierEnum]
or enforce tier to be your enum:
tier: MyTierEnum
```

BryanFauble (approved):
Logic looks great to me, just a few comments.

BWMac:
Thanks for the suggestions @BryanFauble! I made those changes in [ba4d83f](https://github.com/Sage-Bionetworks-Workflows/py-dcqc/pull/56/commits/ba4d83fdff213ad5003e19ce25377c8823823f39) and [ab39fcf](https://github.com/Sage-Bionetworks-Workflows/py-dcqc/pull/56/commits/ab39fcf8784c0692b63ab918e499cfd7d54a1f75).

thomasyu888 (approved):
🔥 LGTM! Excellent work and review!

PR #57, opened 2024-06-17
Pin `requests` version

BWMac:
**Problem:**
There appears to have been a breaking [change](https://github.com/docker/docker-py/issues/3256) for `docker-py` introduced by `requests` v2.32.0. This change should prevent that from breaking our CI and development environments.

**Solution:**
Pin the `requests` version to <=2.31.

**Notes:**
- I also added `gh_token` as an argument to the keepalive-workflow in the CI workflow. See this [PR](https://github.com/Sage-Bionetworks/fs-synapse/pull/4) for the reasoning.
- There was also an additional [issue](https://github.com/Sage-Bionetworks-Workflows/py-dcqc/actions/runs/9568869840/job/26380120004) with the container version being used (deprecation) in the `GrepDateTimeTest`, I bumped it up to the latest published version in `biocontainers`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,8 +45,6 @@ jobs:
         run: |
           echo "wheel=$(ls dist/*.whl)" >> $GITHUB_OUTPUT
           echo "tarball=$(ls dist/*.tar.gz)" >> $GITHUB_OUTPUT
-      - name: get requests version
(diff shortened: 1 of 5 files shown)


thomasyu888 (approved):
🔥 LGTM! I know this is in draft mode but thought i'd review

PR #68, opened 2026-04-15
[DPE-1639] Lock down click, typer and setuptools dependencies

rxu17:
# **Problem:**
Our current versions for the following dependencies are breaking when running the docker image, as well as installing the environment locally:
- typer = 0.7.0
- click = 8.2.1
- setuptools = 82.0.1

See JIRA Ticket for details on the errors encountered and the detailed solutions.
JIRA Ticket: https://sagebionetworks.jira.com/browse/DPE-1639

**Note** Due to the [failing CI tests](https://github.com/Sage-Bionetworks-Workflows/py-dcqc/actions/runs/24682366210/job/72182867900) which were due to the `skipped_tests` having the possibility of being None, it needed a safer method of being appended with `external_tests` otherwise it failed in the unit tests

# **Solution:**
Lock down dependencies to non-breaking versions:
- typer ~= 16.1.0
- setuptools ~=65.0

# **Testing:**
[X] - Locally run through all of the steps via python virtual env
[X] - Rebuild the docker image with the updated `setup.cfg` and run through all of the steps
[X] - Compared that outputs are the same as a successful DCQC run on nextflow 

Nextflow output with original docker image with no profile (can't find this run anymore, ghost run?)
image

Nextflow output (ran locally) with newly built docker image and `test`, `local` and `docker` profile
image

Nextflow output (ran on seqera platform) with old docker image and `test` profile
image

The DCQC results match between the local nextflow run (with the updated docker image) and the nextflow output. It's been difficult to test this out locally without nextflow, I think mainly with the TIFF tests and needing the external tiff docker image to do that part?

### DIFF
--- a/src/dcqc/main.py
+++ b/src/dcqc/main.py
@@ -198,6 +198,7 @@ def qc_file(
     # Prepare suite (skip all external tests)
     suite = SuiteABC.from_target(target, required_tests_maybe, skipped_tests)
     external_tests = [test.type for test in suite.tests if test.is_external_test]
+    skipped_tests = skipped_tests or []
     skipped_tests += external_tests
     suite = SuiteABC.from_target(target, required_tests_maybe, skipped_tests)
 
--- a/README.md
+++ b/README.md
@@ -211,10 +211,8 @@ For testing purposes, there is a [sample input targets file that is used in the 
 Run QC on a single file:
 
 ```bash
-dcqc qc-file data.csv --file-type csv --metadata '{"author": "John Doe"}' --skipped-tests ""
+dcqc qc-file data.csv --file-type csv --metadata '{"author": "John Doe"}'
 ```
-
-Note you must provide `--skipped-tests ""` as a empty string if you have no tests to skip
 
 ### Creating and Running Test Suites
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 # AVOID CHANGING REQUIRES: IT WILL BE UPDATED BY PYSCAFFOLD!
-requires = ["setuptools>=46.1.0", "setuptools_scm[toml]>=5"]
+requires = ["setuptools>=80.9.0,<82.0.0", "setuptools_scm[toml]>=5"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools_scm]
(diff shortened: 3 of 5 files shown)


thomasyu888:
🔥 I almost want to say that the original docker container had an error, because the name "make status amber" actually comes up with status as amber now.

rxu17:
> 🔥 I almost want to say that the original docker container had an error, because the name "make status amber" actually comes up with status as amber now.

Super odd is I added the `test` profile to the docker run with the "old" docker image and now it matches the newly built docker image in making status AMBER. Now I can't find the run that had the `RED` status anymore on Nextflow .... Ghost run.

andrewelamb:
I'm not a huge fan of locking packages down like 

```
typer ~= 16.1.0
setuptools ~=65.0
```
but it seems like the best thing to do for now. We should revisit after upgrading to the new fs-synapse version in the near future.

andrewelamb (approved):

rxu17:
> but it seems like the best thing to do for now. We should revisit after upgrading to the new fs-synapse version in the near future.

Yes I think once we get in your ticket of allowing in click >= 8.3.0 and remove `fs`, then we can stop locking down to a specific `typer`, `click` (since typer and click are so interlinked) and `setuptools`. I kept setuptools the same as what it was in the setup.cfg

src/dcqc/main.py
```diff
@@ -198,6 +198,7 @@ def qc_file(
     # Prepare suite (skip all external tests)
     suite = SuiteABC.from_target(target, required_tests_maybe, skipped_tests)
     external_tests = [test.type for test in suite.tests if test.is_external_test]
+    skipped_tests = skipped_tests or []
```
rxu17, line 201: `skipped_tests` can be `None` if not specified by the user input through the cli for example. But external_tests are always [] so `[] + None` will create the errors seen in the CI.

This line needed to be added due to failing tests with CI run here: https://github.com/Sage-Bionetworks-Workflows/py-dcqc/actions/runs/24682366210

I reran the local dcqc commands and also the nextflow run with the rebuilt docker image: 

Image

Also added tests to help check this

This should also be fine in nf-dcqc because the usual process call is something like `dcqc create-tests --required-tests LibTiffInfoTest --required-tests BioFormatsInfoTest --required-tests OmeXmlSchemaTest "target-0001.json" tests/` so `--skipped-tests` is omitted. I tested the run with `--skipped_tests` specified as well and it ran through successfully.

rxu17: cc @andrewelamb @thomasyu888

thomasyu888 (approved):
🔥 LGTM! Thanks for the work

PR #72, opened 2026-05-21
[DPE-1571] Updated FS-Synapse, other old packages and old github actions

andrewelamb:
# **Problem:**

The package depended on [PyFilesystem](https://github.com/PyFilesystem/pyfilesystem2) (`fs~=2.4` plus `fs-synapse~=1.0/2.x`) for all file access. PyFilesystem is effectively unmaintained, which forced a cascade of version pins that blocked the project from running on modern Python and modern dependencies:

- **Stuck on old Python.** `python_requires` was pinned to `>=3.10, <3.12`, and CI only exercised 3.10/3.11. Tracking tickets (ORCA-346, ORCA-352) had accumulated to sunset 3.10 and allow newer interpreters.
- **Stuck on old dependencies.** `requests<=2.31.0`, `click<=8.2.1`, `mypy~=0.9` (v0.991), `pytest~=7`, `hypothesis~=4`, `docker~=6.1`, `pre-commit~=2`, `isort 6` were all held back, again with TODO tickets (ORCA-350, ORCA-351, ORCA-355) noting the locks could not be lifted while `fs` was in the dependency tree. `setuptools~=65` had to be carried only because `fs` calls the deprecated `pkg_resources`.

# **Solution:**

Migrate the filesystem layer from PyFilesystem to [`fsspec`](https://filesystem-spec.readthedocs.io/) (paired with `fs-synapse>=3.0.2`, which is now an fsspec-based implementation). With `fs` removed from the tree, the chain of version pins it forced could be unwound.

**Filesystem migration (`file.py`, `reports.py`):**
- `FS` → `fsspec.spec.AbstractFileSystem`; `open_parent_fs(url)` → `fsspec.url_to_fs(url)`.
- API translations: `getinfo()` → `info()` (dict-based), `fs.download(...)` → `fs.get(...)`, `makedirs(recreate=True)` → `makedirs(exist_ok=True)`, `info.is_dir` → `info["type"] != "directory"`, and `fs.errors.ResourceNotFound` → the standard `FileNotFoundError`.
- `local_path` now uses `Path(self.fs_path)` directly instead of `getsyspath(...)`, reusing the cached `fs_path` and no longer dereferencing symlinks.

**Dependency / toolchain modernization:**
- `python_requires` → `>=3.11, <3.15`; CI matrix and `tox.ini` now cover 3.11–3.14 (drops 3.9/3.10).
- Unpinned/raised: `requests>=2.24.0,<3.0.0` (2.24.0 is the real floor avoiding the 2.22/2.23 security issues), `typer>=0.25.1` with `click>=8.0` now declared explicitly (newer typer no longer pulls click transitively), `mypy~=2.0`, `isort 8.0.1`, `pytest>=9`, `pytest-cov~=7`, `hypothesis~=6`, `docker~=7.1`, `pre-commit~=4`. Removed the `fs`-only `setuptools~=65` pin and the 3.10-only `exceptiongroup`/`tomli` test deps.

This resolves the cluster of ORCA-346/350/351/352/355 TODOs. The main design decision was to adopt `fsspec` as the standard, actively-maintained successor to PyFilesystem rather than fork or pin `fs` indefinitely.

### DIFF
--- a/src/dcqc/file.py
+++ b/src/dcqc/file.py
@@ -224,12 +224,12 @@ class File(SerializableMixin):
                 staged yet and thus has no local copy.
 
         Returns:
-            The path to the local copy or `None` if unavailable.
+            The path to the local copy.
         """
         if self._local_path is None and self.is_url_local():
-            # Reuse the cached filesystem path; do not resolve() so that a
-            # symlinked local file keeps the path the caller supplied rather
-            # than its dereferenced target.
+            # Reuse the cached (absolutized) filesystem path; do not resolve()
+            # so that a symlinked local file keeps its symlink path rather than
+            # its dereferenced target.
             self._local_path = Path(self.fs_path)
         if self._local_path is None:
             message = "Local path is unavailable. Use stage() to create a local copy."
--- a/src/dcqc/file.py
+++ b/src/dcqc/file.py
@@ -22,10 +22,11 @@ from tempfile import gettempdir, mkdtemp
 from typing import Any, ClassVar, Optional
 from warnings import warn
 
+import fsspec
 from fsspec.spec import AbstractFileSystem
 
 from dcqc.mixins import SerializableMixin, SerializedObject
-from dcqc.utils import is_url_local, open_parent_fs
+from dcqc.utils import is_url_local
 
 
 @dataclass
@@ -210,7 +211,7 @@ class File(SerializableMixin):
         Returns:
             A file system + basename pair.
         """
-        fs, fs_path = open_parent_fs(self.url)
+        fs, fs_path = fsspec.url_to_fs(self.url)
         self._fs_path = fs_path
         self._fs = fs
         return fs, fs_path
@@ -227,7 +228,7 @@ class File(SerializableMixin):
             The path to the local copy or `None` if unavailable.
         """
         if self._local_path is None and is_url_local(self.url):
-            _, fs_path = open_parent_fs(self.url)
+            _, fs_path = fsspec.url_to_fs(self.url)
             self._local_path = Path(fs_path).resolve()
         if self._local_path is None:
             message = "Local path is unavailable. Use stage() to create a local copy."
--- a/src/dcqc/file.py
+++ b/src/dcqc/file.py
@@ -254,7 +254,8 @@ class File(SerializableMixin):
     def name(self) -> str:
         """The file name according to the file system."""
         if self._name is None:
-            self._name = Path(self.fs_path).name
+            info = self.fs.info(self.fs_path)
+            self._name = Path(info.get("name") or self.fs_path).name
         return self._name
 
     def get_file_type(self) -> FileType:
--- a/src/dcqc/file.py
+++ b/src/dcqc/file.py
@@ -227,7 +227,7 @@ class File(SerializableMixin):
         Returns:
             The path to the local copy or `None` if unavailable.
         """
-        if self._local_path is None and is_url_local(self.url):
+        if self._local_path is None and self.is_url_local():
             _, fs_path = fsspec.url_to_fs(self.url)
             self._local_path = Path(fs_path).resolve()
         if self._local_path is None:
--- a/src/dcqc/file.py
+++ b/src/dcqc/file.py
@@ -254,10 +254,6 @@ class File(SerializableMixin):
         """The file name according to the file system."""
         if self._name is None:
             info = self.fs.info(self.fs_path)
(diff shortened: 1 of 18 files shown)


src/dcqc/utils.py
```diff
@@ -1,45 +1,37 @@
...
-
-    # Handle the case when there is no "/" in the path
-    if path == "":
-        path = fs_root
-        fs_root = ""
-
-    fs_url = prefix + fs_root
-    fs = open_fs(fs_url)
-    return fs, path
+    return bool(url) and (url.startswith("file://") or "://" not in url)
```
thomasyu888, line 21: I understand everything except I'm getting caught up on the `bool(url)`, can you speak to the importance of that logic?

src/dcqc/reports.py
```diff
@@ -32,11 +31,11 @@ def _create_parent_directories(self, url: str):
...
         fs, fs_path = self._init_fs(parent_url)
         try:
-            info = fs.getinfo(fs_path)
-        except ResourceNotFound:
-            fs.makedirs(fs_path, recreate=True)
-            info = fs.getinfo(fs_path)
-        if not info.is_dir:
+            info = fs.info(fs_path)
+        except FileNotFoundError:
+            fs.makedirs(fs_path, exist_ok=True)
```
thomasyu888, line 36: NIce usage of base python exception

src/dcqc/file.py
```diff
@@ -394,7 +397,8 @@ def stage(
             destination.symlink_to(self._local_path.resolve())
         else:
             with destination.open("wb") as dest_file:
-                self.fs.download(self.fs_path, dest_file)
+                with self.fs.open(self.fs_path, "rb") as src:
+                    shutil.copyfileobj(src, dest_file)
```
thomasyu888: Im not as knowledgeable about `fs` and what is your rationale between the get function https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.get and what you've written here?

thomasyu888 (commented):
I know this is in draft mode, and for the most part, it looks great.  I'll defer to someone else on the team for final review after it's in a none-draft state, and I added comments.

src/dcqc/utils.py
andrewelamb: It's to make sure things like empty strings return as False

src/dcqc/file.py
andrewelamb: It seems that there are trade-offs to using `get` vs `open` but `get` would probably be better here.

andrewelamb:
> I know this is in draft mode, and for the most part, it looks great. I'll defer to someone else on the team for final review after it's in a none-draft state, and I added comments.

 I wanted to get a draft up to see if the GitHub actions worked. It still needs a good amount of review by me, so let's hold off on getting anyone else to review yet. I appreciate the comments though

src/dcqc/utils.py
```diff
@@ -1,45 +1,19 @@
...
-
-
 def is_url_local(url: str) -> bool:
     """Check whether a URL refers to a local location.
 
+    A URL is considered local if it uses the file:// scheme or contains
+    no scheme separator (://), which covers bare absolute and relative paths.
+    Any other scheme (s3://, memory://, syn://, etc.) is treated as remote.
+
+    Only the empty-authority form file:///path is supported. A file:// URL
```
rxu17: I'm a bit confused by this docstring

```
A URL is considered local if it uses the file:// scheme
```
and then it says

```
Only the empty-authority form file:///path is supported. A file:// URL
    with a non-empty authority (e.g. file://host/path) is still reported as
    local here, but fsspec cannot resolve it to a correct on-disk path, so
    such URLs are unsupported.
```
But the code seemed to reflect that `file://...` is considered local but it's unsupported later in the codebase elsewhere then? Is that my understanding?

andrewelamb: You're right; this is not worded well. Here's what's going on: fsspec doesn't handle URLs with non-empty authorities: `file://host/path`.

I reworded this docstring but after thinking about i decided to add a validator to the `File` object constructor that errors out when these types of URLs are given, so now downstream functions (like this one) don't need to worry about them.

tests/test_file.py
rxu17, line 1: Do we need tests for the fallback behavior of the file name resolving to `Path(info.get("name")` or `self.fs_path).name` if `info.get("synapse_entity_name")` and/or one of the others is not available

rxu17 (approved):
LGTM! Just a comment

PR #73, opened 2026-06-19
[DPE-1743] Updated Black package version

andrewelamb:
# **Problem:**
The black package was out of date and was causing a Dependabot issue.

# **Solution:**
Black was updated to the newest version

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,7 @@ repos:
   - id: isort
 
 - repo: https://github.com/psf/black
-  rev: 25.1.0
+  rev: 26.5.1
   hooks:
   - id: black
     language_version: python3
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -28,7 +28,7 @@ repos:
   - id: isort
 
 - repo: https://github.com/psf/black
```
linglp, line 30: if black version gets updated here, do we need to rerun all the files?

andrewelamb: I did it locally, and no files were changed. I was surprised myself; that's why this was a separate PR. I expected a lot of changes.

rxu17 (approved):
LGTM!

.pre-commit-config.yaml
linglp: Okay. Thanks for checking. I could approve now

linglp (approved):
LGTM!

